PR #9, opened 2021-03-15
multipass: introduce Multipass and MultipassError

cjp256:
- Introduce Multipass, a class to wrap the functionality
  of the multipass command-line executable.

- Introduce MultipassError, an exception class for all
  Multipass provider errors.

- Include unit and integration tests. Drop placeholder
  test_todo module.

- Add pytest-subprocess to requirements-dev for use in
  the unit tests.

-----


cjp256:
The integration tests can be run locally for now until I have a spread enabled for this repo.

They also may fail atm because of store issues fwiw...

craft_providers/multipass/multipass.py
```diff
@@ -0,0 +1,381 @@
...
+import json
+import logging
+import pathlib
+import shlex
+import subprocess
+from typing import Any, Dict, List, Optional
+
+from craft_providers import errors
+
+from .errors import MultipassError
```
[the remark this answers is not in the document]
cjp256: Yes isort groups relative imports separately by default.  We can add `no_lines_before = ["LOCALFOLDER"]` to the config to prevent that, if you'd like.  I like the separate grouping, but don't mind either way.

craft_providers/multipass/multipass.py
```diff
@@ -0,0 +1,381 @@
...
+        *,
+        instance_name: str,
+        image: str,
+        cpus: str = None,
+        mem: str = None,
+        disk: str = None,
+    ) -> None:
+        """Launch multipass VM.
+
+        :param instance_name: The name the launched instance_name will have.
```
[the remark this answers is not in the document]
cjp256: Bah, that find/replace caused me more headaches.... will fix.

requirements-dev.txt
```diff
@@ -11,6 +11,7 @@ pydocstyle
 pylint
 pytest
 pytest-runner
+pytest-subprocess
```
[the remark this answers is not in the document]
cjp256: I figured we'd freeze it once we get closer to our first tag.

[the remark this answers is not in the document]
cjp256: I can freeze in another PR if you'd prefer to do it sooner :D

sergiusens:
I just worry that things will start to break due to moving dependencies. A
new major version can come in and cause pain

requirements-dev.txt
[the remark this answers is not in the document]
cjp256: Frozen (and fixed missing requirement)

tests/integration/multipass/conftest.py
```diff
@@ -46,9 +46,9 @@ def instance_name():
 def tmp_instance(
     *,
     instance_name=instance_name,
-    image_name="snapcraft:core20",
+    image_name="20.04",
```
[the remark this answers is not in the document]
cjp256: Apparently we need to give multipass some time.  @saviq said there's a bug he's working on, so I pushed a 20 second sleep (his recommendation) to address.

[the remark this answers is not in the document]
cjp256: https://github.com/canonical/multipass/issues/1995

craft_providers/multipass/multipass.py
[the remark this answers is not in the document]
sergiusens: this is fine, I am fine with automation :-)

sergiusens (approved):
This is a very light weight class, as discussed in the big draft, maybe not even worth it as a class. I would suggest staticmethods but I see the multipass executable path being part of this.

I think this is generally fine, I would however consider the transfer method and pass in instance_name and destination path as two different entities even if internally it is one long string (and probably only when using the multipass cli)

cjp256:
> This is a very light weight class, as discussed in the big draft, maybe not even worth it as a class. I would suggest staticmethods but I see the multipass executable path being part of this.
> 

I did that as an experiment and IIRC I thought it looked a bit ugly. It forced the MultipassInstance to worry about the multipass executable path, a detail which Multipass() is meant to hide away.  OTOH, we could drop the path altogether and rely on "multipass" as determined by the PATH.  That would certainly tidy that up.

It wouldn't be a significant amount of effort to make this change, but now would be the time to do it if anyone feels strongly about it :D

> I think this is generally fine, I would however consider the transfer method and pass in instance_name and destination path as two different entities even if internally it is one long string (and probably only when using the multipass cli)

Yeah, I tried balancing the line of pythonizing vs. consistency with the CLI arguments.  In this case I opted to keep Multipass more closely aligned with the CLI interface, because the instance could be referenced in the `source` or `destination`.  Note that most parameters for this class are strings rather than native python types, though I see I missed one or two.  I tried to minimize parsing output to _intent_ (e.g. list() returns just a list of VM names).

craft_providers/multipass/multipass.py
```diff
@@ -0,0 +1,382 @@
...
+
+    :param multipass_path: Path to multipass command to use.
+    """
+
+    def __init__(
+        self, *, multipass_path: pathlib.Path = pathlib.Path("multipass")
+    ) -> None:
+        self.multipass_path = multipass_path
+
+    def _run(  # pylint: disable=redefined-builtin
```
facundobatista, line 47: Which builtin is being redefined?

craft_providers/multipass/multipass.py
```diff
@@ -0,0 +1,382 @@
...
+        command: List[str],
+        *,
+        check=True,
+        **kwargs,
+    ) -> subprocess.CompletedProcess:
+        """Execute command in instance_name, allowing output to console."""
+        command = [str(self.multipass_path), *command]
+
+        logger.debug("Executing on host: %s", shlex.join(command))
+        return subprocess.run(command, check=check, **kwargs)
```
facundobatista, line 58: If `check` is not used in other place, what's the point in separating it from `kwargs`?

craft_providers/multipass/multipass.py
```diff
@@ -0,0 +1,382 @@
...
+                brief=f"Failed to delete VM {instance_name!r}.",
+                details=errors.details_from_called_process_error(error),
+            ) from error
+
+    def exec(
+        self,
+        *,
+        command: List[str],
+        instance_name: str,
+        runner=subprocess.run,
```
facundobatista, line 86: This one is not documented below. What is the purpose of it?

craft_providers/multipass/multipass.py
```diff
@@ -0,0 +1,382 @@
...
+
+        :raises MultipassError: On error.
+        """
+        command = ["info", instance_name, "--format", "json"]
+
+        try:
+            proc = self._run(command, capture_output=True, check=True, text=True)
+        except subprocess.CalledProcessError as error:
+            if "does not exist" in error.stderr:
+                return None
```
facundobatista, line 116: Why are we hiding this error?

craft_providers/multipass/multipass.py
```diff
@@ -0,0 +1,382 @@
...
+
+        try:
+            proc = self._run(command, capture_output=True, check=True, text=True)
+        except subprocess.CalledProcessError as error:
+            raise MultipassError(
+                brief="Failed to query list of VMs.",
+                details=errors.details_from_called_process_error(error),
+            ) from error
+
+        data_list = json.loads(proc.stdout).get("list", dict())
```
facundobatista, line 177: The default if `"list"` is not present should be a list, right? Not a dict.

craft_providers/multipass/multipass.py
```diff
@@ -0,0 +1,382 @@
...
+        :param delay_mins: Delay shutdown for specified number of minutes.
+
+        :raises MultipassError: on error.
+        """
+        command = ["stop"]
+
+        if delay_mins != 0:
+            command.extend(["--time", str(delay_mins)])
+
+        command.append(instance_name)
```
facundobatista, line 250: Why not include the instance name in the `command` definition above (as other methods do)?

craft_providers/multipass/multipass.py
```diff
@@ -0,0 +1,382 @@
...
+    def transfer_destination_io(
+        self, *, source: str, destination: io.BufferedIOBase, chunk_size: int = 4096
+    ) -> None:
+        """Transfer from source file to destination IO.
+
+        Note that this can't use std{in,out}=open(...) due to LP #1849753.
+
+        :param source: The source path, prefixed with <name:> for a path inside
+            the instance.
+        :param destination: An IO stream to write to.
```
facundobatista, line 289: Missing "chunk" here.

craft_providers/multipass/multipass.py
```diff
@@ -0,0 +1,382 @@
...
+            if not data:
+                break
+
+            destination.write(data)
+
+        try:
+            _, stderr = proc.communicate(timeout=30)
+        except subprocess.TimeoutExpired:
+            proc.kill()
+            _, stderr = proc.communicate()
```
facundobatista, line 314: How this `communicate` work after the process was killed?

craft_providers/multipass/multipass.py
```diff
@@ -0,0 +1,382 @@
...
+    def transfer_source_io(
+        self, *, source: io.BufferedIOBase, destination: str, chunk_size: int = 4096
+    ) -> None:
+        """Transfer to destination path with source IO.
+
+        Note that this can't use std{in,out}=open(...) due to LP #1849753.
+
+        :param source: An IO stream to read from.
+        :param destination: The destination path, prefixed with <name:> for a
+            path inside the instance.
```
facundobatista, line 333: Missing "chunk"

craft_providers/multipass/multipass.py
```diff
@@ -0,0 +1,382 @@
...
+                    cmd=command, stderr=stderr, returncode=proc.returncode
+                ),
+            )
+
+    def umount(self, *, mount: str) -> None:
+        """Unmount target in VM.
+
+        :param mount: Mount point in <name>[:<path>] format, where <name> are
+            instance names, and optional <path> are mount points.  If omitted,
+            all mounts will be removed from the name instance.
```
facundobatista, line 370: "from the named instance" or "from the <name> instance"?

requirements-dev.txt
```diff
@@ -1,20 +1,79 @@
...
+tox==3.23.0
+tqdm==4.59.0
+twine==3.4.1
+typed-ast==1.4.2
+typing-extensions==3.7.4.3
+urllib3==1.26.4
+virtualenv==20.4.3
+webencodings==0.5.1
+wrapt==1.12.1
+zipp==3.4.1
```
facundobatista, line 79: Let's talk about this. I don't like it.

My main problem is: how do you ever get out from these fixed dependencies versions and evolve?

tests/integration/multipass/test_multipass.py
```diff
@@ -0,0 +1,244 @@
...
+import subprocess
+
+import pytest
+
+from craft_providers.multipass import Multipass
+
+from . import conftest
+
+pytestmark = pytest.mark.skipif(
+    shutil.which("multipass") is None, reason="multipass not installed"
```
facundobatista, line 26: Shouldn't we force the developer to have multipass installed and always test these?

facundobatista (changes requested):
Great work! I annotated some details and something we need to talk at a team level.

tests/integration/multipass/test_multipass.py
cjp256: My thoughts:
- not all providers will support all platforms (e.g. lxd on Windows)
- not all hosts running the tests will have the capability to support the provider (e.g. Ubuntu 20.04 host without virtualization capabilities, or already running in an LXC container).

By simply keying off whether or not the developer installed it, we can reasonably determine which tests to skip automatically with the given rationale.  It's easy enough for the developer to determine, and doesn't require any magic knowledge in the test logic.

craft_providers/multipass/multipass.py
cjp256: Best practice according to docs.
https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate

craft_providers/multipass/multipass.py
cjp256: Good catch, from an old name. Removed

craft_providers/multipass/multipass.py
cjp256: check is defaulted to True. While I try to be explicit about check all of the time, it's a safe(r) default and avoids the pylint complaint of:
`craft_providers/multipass/multipass.py:56:15: W1510: Using subprocess.run without explicitly set `check` is not recommended. (subprocess-run-check)`  If there's an easy to punt that check, I'd be interested :D

craft_providers/multipass/multipass.py
cjp256: fixed docs, allows you to use Popen or whatever other subprocess-y function to invoke.

craft_providers/multipass/multipass.py
cjp256: It was the approach used to determine if the VM exists...

However, on second thought, I will rewrite that logic to rely on list().

Dropped.

craft_providers/multipass/multipass.py
cjp256: Yes indeed, though never quite happens.  Just a sane fallback.
```
$ multipass list --format=json
{
    "list": [
    ]
}
```

cjp256: fixed

craft_providers/multipass/multipass.py
cjp256: Fixed

craft_providers/multipass/multipass.py
cjp256: Fixed

requirements-dev.txt
cjp256: I reverted the requirements freeze for now. I agree we need a better approach.  I've seen some projects maintain two lists, one with the actual requirements unpinned, another with the frozen.

craft_providers/multipass/multipass.py
cjp256: Fair question, I fixed the ordering :D

craft_providers/multipass/multipass.py
cjp256: Fixed

tests/integration/multipass/conftest.py
```diff
@@ -0,0 +1,99 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Fixtures for MULTIPASS integration tests."""
```
cmatsuoka, line 15: Multipass written in all caps is unusual.

cjp256, line 15:
```suggestion
"""Fixtures for Multipass integration tests."""
```

cjp256: Fixed

craft_providers/multipass/multipass.py
facundobatista: As we discussed in mattermost, this will be replaced/simplified in the next branch.

facundobatista (approved):
Approving, but please improve/fix the stuff in both comments in the next branch.

PR #13, opened 2021-03-31
util: introduce util package with os_release parser

cjp256:

-----


sergiusens (approved):

craft_providers/util/os_release.py
```diff
@@ -0,0 +1,53 @@
...
+
+    :param content: String contents of os-release file.
+
+    :returns: Dictionary of key-mappings found in os-release. Values are
+              stripped of encapsulating double-quotes.
+
+    """
+    mappings: Dict[str, str] = dict()
+
+    for line in content.splitlines():
```
facundobatista, line 45: Consider using `parse_keqv_list(parse_http_list(content.splitlines()))` (I need to bring these functions to my code too). Both functions can be imported from `urllib.request`.

tests/unit/util/test_os_release.py
```diff
@@ -0,0 +1,62 @@
...
+                "SUPPORT_URL": "https://help.ubuntu.com/",
+                "UBUNTU_CODENAME": "groovy",
+                "VERSION": "20.10 (Groovy Gorilla)",
+                "VERSION_CODENAME": "groovy",
+                "VERSION_ID": "20.10",
+            },
+        ),
+        ('foo="quotes-to-remove"', {"foo": "quotes-to-remove"}),
+        ("foo=extra=equal", {"foo": "extra=equal"}),
+        ('foo="quote-in-"the-middle"', {"foo": 'quote-in-"the-middle'}),
```
facundobatista, line 58: I'll add a test case like `foo=quote-the-"last-part"`, which should produce `{'foo': 'quote-the-"last-part"'}`; note that I think you already support it.

facundobatista (approved):
Approving, but please check the comments.

craft_providers/util/os_release.py
cjp256: I feel like this might reduce code but adds a bit of hidden complexity.

I improved spec conformance and added a bunch more tests.

tests/unit/util/test_os_release.py
cjp256: In the same spirit, i also added "start" tests.

PR #14, opened 2021-04-02
multipass, util: introduce util.env_cmd module

cjp256:


craft_providers/util/env_cmd.py
```diff
@@ -0,0 +1,50 @@
...
+    :param ignore_environment: Start with an empty environment.
+
+    :returns: List of env command strings.
+    """
+    env_cmd = ["env"]
+
+    if ignore_environment:
+        env_cmd.append("-i")
+
+    if env is None:
```
facundobatista, line 41: Does it make sense to ignore the environment, and also set env vars?

I assume it doesn't. What do you think  about `assert env is None` in the block above, and directly return env_cmd without ending the `if`? That way we cover for a potential misuse of the function by mistake.

facundobatista (changes requested):
Looks good! Added small comment.

craft_providers/util/env_cmd.py
cjp256: It does if you want env to mask the default environment and initialize env with only with what you specify.  I've personally used it in a number of cases in the past.

Whether or not we'll use it here, I am unsure.  I just added it for completeness should we want to.

cjp256:
Rebased, and dropped the empty "env" usage if env is None.

craft_providers/util/env_cmd.py
facundobatista: Ah, it ignores the *previous* environment. Ok.

facundobatista (approved):
Awesome!

sergiusens (approved):

PR #15, opened 2021-04-02
bases: introduce abstract Base and buildd implementation

cjp256:
Depends on #13 and #14.

### DIFF
--- a/craft_providers/base.py
+++ b/craft_providers/base.py
@@ -1,0 +1,61 @@
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Base configuration module."""
+import logging
+from abc import ABC, abstractmethod
+from typing import Optional
+
+from .executor import Executor
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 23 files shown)


craft_providers/base.py
```diff
@@ -0,0 +1,61 @@
...
+from abc import ABC, abstractmethod
+from typing import Optional
+
+from .executor import Executor
+
+logger = logging.getLogger(__name__)
+
+
+class Base(ABC):
+    """Base definition for bootstrap configuration and setup.
```
sergiusens, line 26: This docstring does not really tell me much

craft_providers/base.py
```diff
@@ -0,0 +1,61 @@
...
+
+    :param name: Name of base.
+    """
+
+    def __init__(self, *, name: str) -> None:
+        self.name = name
+
+    @abstractmethod
+    def setup(self, *, executor: Executor) -> None:
+        """Configure instance as necessary.
```
sergiusens, line 36:
```suggestion
        """Configure instance with the necessary environment.
```

But it begs the question, what is the necessary environment, or what should classes inheriting from this one implement

craft_providers/bases/craft_config.py
```diff
@@ -0,0 +1,102 @@
...
+from craft_providers import Executor, errors
+
+from .errors import BaseConfigurationError
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass_json
+@dataclass
+class CraftBaseConfig:
```
sergiusens, line 36: can this module be its own PR?

craft_providers/base.py
```diff
@@ -0,0 +1,61 @@
...
+import logging
+from abc import ABC, abstractmethod
+from typing import Optional
+
+from .executor import Executor
+
+logger = logging.getLogger(__name__)
+
+
+class Base(ABC):
```
sergiusens, line 25: can this be its own PR? It would have the benefit of being independent from the other ones too

sergiusens (commented):
I would suggest leaving this PR as reference and splitting it up a bit more

cjp256:
Breaking this up, i first started with Base @ https://github.com/canonical/craft-providers/pull/18.

PR #16, opened 2021-04-02
multipass: add version() and wait_until_ready() to Multipass

cjp256:
wait_until_ready(), strictly speaking, is not a command of
Multipass, but as a helper which is dependent on the Multipass
CLI, seem best fit to include there.  Return the version when
ready.

-----


craft_providers/multipass/multipass.py
```diff
@@ -386,3 +387,65 @@ def umount(self, *, mount: str) -> None:
...
+            if multipassd_version is not None:
+                return (multipass_version, multipassd_version)
+
+            time.sleep(retry_wait)
+
+        raise MultipassError(
+            brief="Timed out waiting for Multipass to become ready.",
+        )
+
+    def version(self):
```
sergiusens, line 411: why no return type?

craft_providers/multipass/multipass.py
```diff
@@ -386,3 +387,65 @@ def umount(self, *, mount: str) -> None:
...
+            - ['multipass', '1.5.0+win', 'multipassd', '1.5.0+win']
+
+        In some cases, Multipass may spew additional data, so protect against
+        it.  See: https://github.com/canonical/multipass/issues/2020
+
+        :returns: Tuple of parsed versions (multipass, multipassd).  multipassd
+                  may be None if Multipass is not yet ready.
+        """
+        try:
+            proc = self._run(["version"], capture_output=True, check=True, text=True)
```
sergiusens, line 431: text=True, was not aware of that one, do we get decode errors if decoding fails?

craft_providers/multipass/multipass.py
```diff
@@ -386,3 +387,65 @@ def umount(self, *, mount: str) -> None:
...
+
+        output_split = proc.stdout.split()
+        if len(output_split) < 2 or output_split[0] != "multipass":
+            raise MultipassError(
+                brief=f"Unable to parse version output: {proc.stdout!r}",
+            )
+
+        multipass_version = output_split[1].split("+")[0]
+
+        if len(output_split) >= 4 and output_split[2] == "multipassd":
```
sergiusens, line 446: code comment with what to expect in multipass_version would be nice

craft_providers/multipass/multipass.py
sergiusens, line 411: return type

sergiusens (changes requested):

craft_providers/multipass/multipass.py
```diff
@@ -386,3 +387,65 @@ def umount(self, *, mount: str) -> None:
...
+        """Wait until Multipass is ready (upon install/startup).
+
+        :param retry_wait: Time to sleep between retries.
+        :param timeout: Optional time.time() stamp to timeout after.
+        """
+        while timeout is None or time.time() < timeout:
+            multipass_version, multipassd_version = self.version()
+
+            if multipassd_version is not None:
+                return (multipass_version, multipassd_version)
```
cmatsuoka, line 403: It's not obvious (not documented) that `wait_until_ready` returns a tuple of versions. Perhaps we could leave that to the `version()` call?

craft_providers/multipass/multipass.py
```diff
@@ -386,3 +387,65 @@ def umount(self, *, mount: str) -> None:
...
+
+            - ['multipass', '1.5.0+mac', 'multipassd', '1.5.0+mac']
+
+            - ['multipass', '1.5.0+win', 'multipassd', '1.5.0+win']
+
+        In some cases, Multipass may spew additional data, so protect against
+        it.  See: https://github.com/canonical/multipass/issues/2020
+
+        :returns: Tuple of parsed versions (multipass, multipassd).  multipassd
+                  may be None if Multipass is not yet ready.
```
cmatsuoka, line 428: It could be useful to return some kind of mapping between component and version, such as a dictionary from component name to version, or a list of (component, version) tuples, e.g. to allow a more uniform return format for the `version()` call for different providers.

craft_providers/multipass/multipass.py
```diff
@@ -386,3 +387,65 @@ def umount(self, *, mount: str) -> None:
...
                 details=errors.details_from_called_process_error(error),
             ) from error
+
+    def wait_until_ready(
+        self, *, retry_wait: float = 0.25, timeout: Optional[float] = None
+    ):
+        """Wait until Multipass is ready (upon install/startup).
+
+        :param retry_wait: Time to sleep between retries.
+        :param timeout: Optional time.time() stamp to timeout after.
```
cmatsuoka, line 397: Specifying an amount of time seems more common than defining a deadline when setting timeout values.

craft_providers/multipass/multipass.py
cjp256: Fixed

craft_providers/multipass/multipass.py
cjp256: Explicitly added decoding failure logic

craft_providers/multipass/multipass.py
cjp256: Fixed.  Will need to revisit with another PR for stricter mypy config.  This should be caught automatically.

craft_providers/multipass/multipass.py
cjp256: I fixed the docs and return annotation.

It's really just to avoid an unnecessary call to version since it will be available. It's almost a half second in some cases.

craft_providers/multipass/multipass.py
cjp256: I don't know if we'll get any other components in the future.  At the moment I don't think we'd care about them.  I'm returning it for completeness and because we use it to identify if multipass is ready.  As it is, the only upcoming usage of it in craft providers ignores the the second (multipassd) value.  How about we keep this for now and I'll revisit in the future when multipass supports a --format=yaml option here (open ticket).

craft_providers/multipass/multipass.py
cjp256: Fixed as well as improved deadline time check..

craft_providers/multipass/multipass.py
cjp256: Also, fwiw this `version()` call is not really exposed to the applications btw.

craft_providers/multipass/multipass.py
```diff
@@ -386,3 +388,105 @@ def umount(self, *, mount: str) -> None:
...
+    def wait_until_ready(
+        self, *, retry_wait: float = 0.25, timeout: Optional[float] = None
+    ) -> Tuple[str, Optional[str]]:
+        """Wait until Multipass is ready (upon install/startup).
+
+        :param retry_wait: Time to sleep between retries.
+        :param timeout: Optional time.time() stamp to timeout after.
+
+        :returns: Tuple of parsed versions (multipass, multipassd).  multipassd
+            may be None if Multipass is not yet ready.
```
sergiusens, line 401:
```suggestion
            may be None if Multipass is not yet and the timeout limit is reached.
```

craft_providers/multipass/multipass.py
```diff
@@ -386,3 +388,105 @@ def umount(self, *, mount: str) -> None:
...
+        """
+        if timeout is not None:
+            deadline: Optional[float] = time.time() + timeout
+        else:
+            deadline = None
+
+        while True:
+            multipass_version, multipassd_version = self.version()
```
sergiusens, line 409: why don't you try and catch this exception here? I guess we can wait and see what happens

cjp256: Which exception is that?  version() may raise MultipassError() which can't be handled (unexpected error cases).

craft_providers/multipass/multipass.py
cmatsuoka: Thanks. But it seems that the timeout parameter description in the docstring still describes a deadline?

cmatsuoka (approved):
LGTM, with the timeout parameter docstring fix.

craft_providers/multipass/multipass.py
```diff
@@ -386,3 +388,105 @@ def umount(self, *, mount: str) -> None:
...
                 details=errors.details_from_called_process_error(error),
             ) from error
+
+    def wait_until_ready(
+        self, *, retry_wait: float = 0.25, timeout: Optional[float] = None
+    ) -> Tuple[str, Optional[str]]:
+        """Wait until Multipass is ready (upon install/startup).
+
+        :param retry_wait: Time to sleep between retries.
+        :param timeout: Optional time.time() stamp to timeout after.
```
cjp256, line 398:
```suggestion
        :param timeout: Optional timeout in seconds.
```

craft_providers/multipass/multipass.py
cjp256: Indeed, forget to fix it here. Fixed!

sergiusens (approved):

craft_providers/multipass/multipass.py
```diff
@@ -386,3 +388,105 @@ def umount(self, *, mount: str) -> None:
...
+    def wait_until_ready(
+        self, *, retry_wait: float = 0.25, timeout: Optional[float] = None
+    ) -> Tuple[str, Optional[str]]:
+        """Wait until Multipass is ready (upon install/startup).
+
+        :param retry_wait: Time to sleep between retries.
+        :param timeout: Timeout in seconds.
+
+        :returns: Tuple of parsed versions (multipass, multipassd).  multipassd
+            may be None if Multipass is not yet and the timeout limit is reached.
```
sergiusens, line 401:
```suggestion
            may be None if Multipass is not ready and the timeout limit is reached.
```

PR #20, opened 2021-04-07
bases: introduce Buildd bases with support for 16.04, 18.04, and 20.04

cjp256:
The splattered _check_deadlines() is a little ugly, but
deadlines are checked after every non-trivial operation.
Perhaps future work can consider using something like stopit,
but we don't really have any strict requirements for a timeout
at the moment.

Note that 16.04 is not supported on Windows with Multipass.

- Add FakeExecutor fixture to simplify unit tests requiring executors.
  Paired with fake_subprocess, tests have a lot of control.

-----


sergiusens:
> Note that 16.04 is not supported on Windows with Multipass.

It will be in the next Multipass release :-D

craft_providers/bases/buildd.py
```diff
@@ -0,0 +1,520 @@
...
+def default_command_environment() -> Dict[str, Optional[str]]:
+    """Provide default command environment dictionary.
+
+    The minimum environment for the buildd image to be configured and
+    function properly.
+
+    :returns: Dictionary of environment key/values.
+    """
+    return dict(
+        PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin"
```
sergiusens, line 42: please add a comment explaining why this PATH

craft_providers/bases/buildd.py
```diff
@@ -0,0 +1,520 @@
...
+        :param deadline: Optional time.time() deadline.
+        """
+        try:
+            executor.execute_run(
+                [
+                    "apt-get",
+                    "install",
+                    "-y",
+                    "fuse",
+                    "udev",
```
sergiusens, line 347: ask around to see if we still need udev :-)

sergiusens (approved):
A bit long, I might have missed some things, but it also looks pretty familiar to what we have on Snapcraft

craft_providers/bases/buildd.py
cjp256: I think I hit this myself, but I can run a set of tests without it.

cjp256: Actually, I'll put a TODO here to revisit after LXD provider is back in shape.  And now that I see my TODOs all the time, it won't get lost :rofl:

craft_providers/bases/buildd.py
cjp256: done.

craft_providers/bases/buildd.py
```diff
@@ -0,0 +1,522 @@
...
+        :param timeout: Timeout in seconds.
+
+        :raises BaseConfigurationError: on other unexpected error.
+        """
+        if timeout is not None:
+            deadline: Optional[float] = time.time() + timeout
+        else:
+            deadline = None
+
+        self._setup_environment(executor=executor, deadline=deadline)
```
cmatsuoka, line 138: This probably won't make much difference in a real world usage, but it seems that even if the deadline has already been reached it will still execute the first action in `_setup_environment()`. If I start with an expired deadline I would expect it to do nothing at all.

craft_providers/bases/buildd.py
```diff
@@ -0,0 +1,522 @@
...
+                env=self.command_environment,
+            )
+            _check_deadline(deadline)
+        except subprocess.CalledProcessError as error:
+            raise BaseConfigurationError(
+                brief="Failed to install apt-utils.",
+                details=errors.details_from_called_process_error(error),
+            ) from error
+
+        _check_deadline(deadline)
```
cmatsuoka, line 190: Is this last check necessary? It already checked in the previous `try` block.

craft_providers/bases/buildd.py
```diff
@@ -0,0 +1,522 @@
...
+        )
+        self._setup_apt(executor=executor, deadline=deadline)
+        self._setup_snapd(executor=executor, deadline=deadline)
+
+    def _setup_apt(self, *, executor: Executor, deadline: Optional[float]) -> None:
+        """Configure apt & update cache.
+
+        :param executor: Executor for target container.
+        """
+        executor.create_file(
```
cmatsuoka, line 156: Maybe we should check if deadline was met also at the start of each call so that nothing would be executed in that case?

craft_providers/bases/buildd.py
```diff
@@ -0,0 +1,522 @@
...
+                    "fuse",
+                    "udev",
+                ],
+                check=True,
+                capture_output=True,
+                env=self.command_environment,
+            )
+            _check_deadline(deadline)
+
+            executor.execute_run(
```
cmatsuoka, line 357: Since all these calls are the same except for the command being executed, perhaps we could iterate through a sequence of commands?

craft_providers/bases/buildd.py
cjp256: Agreed, fixed.

craft_providers/bases/buildd.py
cjp256: Fixed!

craft_providers/bases/buildd.py
cjp256: They could, but I'd rather just spend a few extra lines and have the flexibility if I need to change a parameter or handling for just one call later.

craft_providers/bases/buildd.py
cjp256: Fixed!

craft_providers/bases/buildd.py
cjp256: To prevent redundant checks, I decided to call deadline checks after each command.  But it always bothered me the last operation might succeed and deadline will raise error.  I moved all checks to the front.

cmatsuoka (approved):
Thanks!

PR #21, opened 2021-04-08
multipass: introduce MultipassProvider

cjp256:
Compatibility and cleaning checks will be introduced later.

Depends on #20

### DIFF
--- a/craft_providers/multipass/multipass_provider.py
+++ b/craft_providers/multipass/multipass_provider.py
@@ -1,0 +1,154 @@
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Multipass Provider."""
+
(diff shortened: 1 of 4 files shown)


craft_providers/multipass/multipass_provider.py
```diff
@@ -0,0 +1,154 @@
...
+    def _configure_instance(
+        self,
+        *,
+        instance: MultipassInstance,
+        base_configuration: Base,
+    ) -> None:
+        """Configure instance.
+
+        Start to instance first to ensure it is started, as well as to cancel
+        any outstanding delay-shtudown request.
```
sergiusens, line 70:
```suggestion
        Start the instance first to ensure it is started, as well as to cancel
        any outstanding delay-shtudown request.
```

craft_providers/multipass/multipass_provider.py
```diff
@@ -0,0 +1,154 @@
...
+                image=image_name,
+            )
+
+        self._configure_instance(
+            instance=instance,
+            base_configuration=base_configuration,
+        )
+        return instance
+
+    def install(self) -> str:
```
sergiusens, line 128: souldn't this be a static class method?

craft_providers/multipass/multipass_provider.py
```diff
@@ -0,0 +1,154 @@
...
+            _install_windows()
+        else:
+            raise errors.MultipassInstallationError(
+                f"unsupported platform {sys.platform!r}"
+            )
+
+        multipass_version, _ = Multipass().wait_until_ready()
+        return multipass_version
+
+    def is_installed(self) -> bool:
```
sergiusens, line 149: souldn't this be a static class method?

craft_providers/multipass/multipass_provider.py
cjp256: Probably should just drop the class altogether.

PR #22, opened 2021-04-09
multipass: introduce installation support

cjp256:
- Add MultipassInstallationError.

- Add unit tests for Multipass errors (formatting).

- Add unit & integration tests for installer bits.

- Update CI to allow MacOS integration tests to uninstall multipass.

-----

### DIFF
--- a/craft_providers/multipass/installer.py
+++ b/craft_providers/multipass/installer.py
@@ -18,6 +18,7 @@ import logging
 import shutil
 import subprocess
 import sys
+import time
 
 from craft_providers.errors import details_from_called_process_error
 
@@ -46,6 +47,12 @@ def install() -> str:
         raise errors.MultipassInstallationError(
             f"unsupported platform {sys.platform!r}"
         )
+
+    # TODO: Multipass needs time after being installed or errors could happen on
+    # launch, i.e.: "Remote "" is unknown or unreachable." Current guidance is
+    # to sleep 20 seconds after install, but we should have a more reliable and
+    # timely approach. See: https://github.com/canonical/multipass/issues/1995
+    time.sleep(20)
 
     multipass_version, _ = Multipass().wait_until_ready()
     return multipass_version
--- a/craft_providers/multipass/installer.py
+++ b/craft_providers/multipass/installer.py
@@ -80,4 +80,4 @@ def is_installed() -> bool:
 
     :returns: Bool if multipass is installed.
     """
-    return not shutil.which("multipass") is None
+    return shutil.which("multipass") is not None
--- a/craft_providers/multipass/installer.py
+++ b/craft_providers/multipass/installer.py
@@ -1,0 +1,83 @@
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Multipass Provider."""
+
+import logging
+import shutil
+import subprocess
+import sys
+
+from craft_providers.errors import details_from_called_process_error
+
+from . import errors
+from .multipass import Multipass
+
+logger = logging.getLogger(__name__)
+
+
+def install() -> str:
+    """Install Multipass.
+
+    :returns: Multipass version.
+
+    :raises MultipassInstallationError: on error.
+    """
+    if sys.platform == "darwin":
+        _install_darwin()
+    elif sys.platform == "linux":
+        _install_linux()
+    elif sys.platform == "win32":
+        _install_windows()
+    else:
+        raise errors.MultipassInstallationError(
+            f"unsupported platform {sys.platform!r}"
+        )
+
+    multipass_version, _ = Multipass().wait_until_ready()
+    return multipass_version
+
+
+def _install_darwin() -> None:
+    try:
+        subprocess.run(["brew", "install", "multipass"], check=True)
+    except subprocess.CalledProcessError as error:
+        raise errors.MultipassInstallationError(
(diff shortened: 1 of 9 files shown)


tests/unit/multipass/test_installer.py
```diff
@@ -0,0 +1,119 @@
...
+        ["multipass", "version"],
+    ]
+
+
+def test_install_linux_error(
+    fake_process, mock_details_from_process_error, monkeypatch
+):
+    monkeypatch.setattr(sys, "platform", "linux")
+    fake_process.register_subprocess(
+        ["sudo", "snap", "install", "multipass"],
```
sergiusens, line 89: I wonder if we should just remove sudo here

cjp256: The `snap` command won't install without root?  I expect the application will interact with the user before invoking install().

cjp256:
Apologies for the forced push.

tests/unit/multipass/test_installer.py
sergiusens: most systems will trigger a snap install over policy kit

craft_providers/multipass/installer.py
```diff
@@ -0,0 +1,83 @@
...
+        subprocess.run(["sudo", "snap", "install", "multipass"], check=True)
+    except subprocess.CalledProcessError as error:
+        raise errors.MultipassInstallationError(
+            "error during snap installation",
+            details=details_from_called_process_error(error),
+        ) from error
+
+
+def _install_windows() -> None:
+    raise errors.MultipassInstallationError(
```
facundobatista, line 73: I'd simplify the code and just raise this error above in the `if-elif` structure...

craft_providers/multipass/installer.py
```diff
@@ -0,0 +1,83 @@
...
+        "automated installation not yet supported for Windows"
+    )
+
+
+def is_installed() -> bool:
+    """Check if Multipass is installed (and found on PATH).
+
+    :returns: Bool if multipass is installed.
+    """
+    return not shutil.which("multipass") is None
```
facundobatista, line 83: the used idiom is `return shutil.which("multipass") is not None`

tests/integration/multipass/conftest.py
```diff
@@ -97,3 +101,43 @@ def tmp_instance(
...
+
+@pytest.fixture(autouse=True, scope="module")
+def installed_multipass():
+    """Ensure multipass is installed, or skip the test if we cannot.
+
+    If the environment has CRAFT_PROVIDERS_TESTS_ENABLE_MULTIPASS_INSTALL=1,
+    force the installation of Multipass if uninstalled.
+    """
+    if os.environ.get("CRAFT_PROVIDERS_TESTS_ENABLE_MULTIPASS_INSTALL") == "1":
+        multipass.install()
```
facundobatista, line 114: It's really scary to run tests and those tests use sudo and ask you for your password.

Even if one set an envvar (which may not fully understand).

Can't we go to the model of "using multipass if present, skipping if not, in this case telling the developer to check the README for installation instructions"?

tests/integration/multipass/test_installer.py
```diff
@@ -0,0 +1,32 @@
...
+def test_is_installed():
+    expected = shutil.which("multipass") is not None
+
+    assert multipass.is_installed() is expected
+
+
+def test_install(uninstalled_multipass):  # pylint: disable=unused-argument
+    assert multipass.is_installed() is False
+
+    multipass_version = multipass.install()
```
facundobatista, line 29: This really installs multipass in the system as part of the test?

Wouldn't this mess with the system too much?

facundobatista (changes requested):
Nice! Added some questions inline, thanks!

tests/integration/multipass/test_installer.py
sergiusens: this runs under spread by default... usually all our integration tests are dangerous on native systems which is why we had them completely separate (as spread tasks)

tests/integration/multipass/conftest.py
cjp256: It defaults to that behavior (see below elif).

I cannot imagine anyone would honestly be surprised that it tried to install Multipass when setting CRAFT_PROVIDERS_TESTS_ENABLE_MULTIPASS_INSTALL and attempting to run the Multipass integration tests.

But I'll remove it, unless you can think of a foolproof envvar alternative :rofl:

tests/integration/multipass/test_installer.py
cjp256: No.  It's an integration test, how else are we supposed to get automated coverage for it? :P

And for safety, it's only uninstalled if you specifically set the envvar to do so (CRAFT_PROVIDERS_TESTS_ENABLE_MULTIPASS_UNINSTALL=1)

tests/unit/multipass/test_installer.py
cjp256: Ah, you are indeed correct.  I do not have the polkit agent running so it never happened for me before :D  But I started it and it worked as you'd expect.

craft_providers/multipass/installer.py
cjp256: yes indeed, fixed!

craft_providers/multipass/installer.py
cjp256: sure, done.

craft_providers/multipass/installer.py
```diff
@@ -0,0 +1,79 @@
...
+        raise errors.MultipassInstallationError(
+            "error during snap installation",
+            details=details_from_called_process_error(error),
+        ) from error
+
+
+def is_installed() -> bool:
+    """Check if Multipass is installed (and found on PATH).
+
+    :returns: Bool if multipass is installed.
```
cmatsuoka, line 77: Perhaps we could remove the Bool from the description? This can be interpreted as "it returns a bool only if multipass is installed" (or maybe you meant True instead of Bool?)

sergiusens (approved):

tests/integration/multipass/conftest.py
facundobatista: Ah! Here (and below) I misread/confused unit with integration tests, sorry!

facundobatista (approved):
Great, thanks!!

craft_providers/multipass/installer.py
cjp256: Fixed

PR #23, opened 2021-04-09
multipass: introduce launch to create and configure instances

cjp256:
Depends on #22

### DIFF
--- a/craft_providers/multipass/_launch.py
+++ b/craft_providers/multipass/_launch.py
@@ -27,18 +27,18 @@ def launch(
     name: str,
     *,
     base_configuration: Base,
+    image_name: str,
     cpus: int = 2,
     disk_gb: int = 64,
-    image_name: str,
     mem_gb: int = 2,
 ) -> MultipassInstance:
     """Create, start, and configure instance.
 
     :param name: Name of instance.
     :param base_configuration: Base configuration to apply to instance.
+    :param image_name: Multipass image to use, e.g. snapcraft:core20.
     :param cpus: Number of CPUs.
     :param disk_gb: Disk allocation in gigabytes.
-    :param image_name: Multipass image to use, e.g. snapcraft:core20.
     :param mem_gb: Memory allocation in gigabytes.
 
     :returns: Multipass instance.
@@ -49,6 +49,7 @@ def launch(
     instance = MultipassInstance(name=name)
 
     if instance.exists():
+        # TODO: Warn if existing instance doesn't match cpu/disk/mem specs.
         instance.start()
     else:
         instance.launch(
--- a/craft_providers/multipass/_launch.py
+++ b/craft_providers/multipass/_launch.py
@@ -28,7 +28,7 @@ def launch(
     *,
     base_configuration: Base,
     cpus: int = 2,
-    disk_gb: int = 256,
+    disk_gb: int = 64,
     image_name: str,
     mem_gb: int = 2,
 ) -> MultipassInstance:
--- a/craft_providers/multipass/_launch.py
+++ b/craft_providers/multipass/_launch.py
@@ -1,0 +1,62 @@
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


craft_providers/multipass/_launch.py
```diff
@@ -0,0 +1,62 @@
...
+    image_name: str,
+    mem_gb: int = 2,
+) -> MultipassInstance:
+    """Create, start, and configure instance.
+
+    :param name: Name of instance.
+    :param base_configuration: Base configuration to apply to instance.
+    :param cpus: Number of CPUs.
+    :param disk_gb: Disk allocation in gigabytes.
+    :param image_name: Multipass image to use, e.g. snapcraft:core20.
```
sergiusens, line 41: @saviq can we have a "craft" alias in a near future?

craft_providers/multipass/_launch.py
```diff
@@ -0,0 +1,62 @@
...
+    :param mem_gb: Memory allocation in gigabytes.
```
sergiusens, line 42: is ordering alphabetically better here than by context/domain?

craft_providers/multipass/_launch.py
```diff
@@ -0,0 +1,62 @@
...
+
+    :returns: Multipass instance.
+
+    :raises BaseConfigurationError: on unexpected error configuration base.
+    :raises MultipassError: on unexpected Multipass error.
+    """
+    instance = MultipassInstance(name=name)
+
+    if instance.exists():
```
sergiusens, line 51: for future setups, maybe warn that the instance that exists does not satisfy the requested CPU, Disk or RAM

craft_providers/multipass/_launch.py
Saviq: So that `snapcraft:*` equals `craft:*`? Which one would we publicize going forward?

tests/integration/multipass/test_launch.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+@pytest.mark.parametrize(
+    "alias,image_name",
+    [
+        pytest.param(
+            BuilddBaseAlias.XENIAL,
+            "snapcraft:core",
+            marks=pytest.mark.skipif(
+                sys.platform == "win32", reason="unsupported on windows"
```
sergiusens, line 42: this might not be the case very soon

craft_providers/multipass/_launch.py
cjp256: Good idea.

craft_providers/multipass/_launch.py
```diff
@@ -0,0 +1,62 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def launch(
+    name: str,
+    *,
+    base_configuration: Base,
+    cpus: int = 2,
+    disk_gb: int = 256,
```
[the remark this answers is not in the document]
cjp256: 256G was the default in Snapcraft.  I'd think 5G would definitely hit issues. 

The multipass disks are sparse so it shouldn't be significantly different on disk.  I'd rather lean towards having a bit too much to prevent people from having to figure out how to configure it.  But I agree perhaps 256G is a bit too much if the user uses it often and never cleans.

@sergiusens Do you by chance have any data on a "large project" built by Snapcraft?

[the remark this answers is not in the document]
cjp256: Launchpad has a 60G quota.  So I'll change this 64G for now :D

tests/integration/multipass/test_launch.py
cjp256: Added TODO

craft_providers/multipass/_launch.py
cjp256: I'm indifferent, but reordered :D

craft_providers/multipass/_launch.py
cjp256: Added TODO

craft_providers/multipass/_launch.py
sergiusens: We are going to use this for our other craft implementations (e.g.; charmcraft), so I guess calling it craft will have a longer run than other names.

We really don't expose the name unless there is an error, I suspect it will be confusing to say "Failed to launch snapcraft:core20" in a charmcraft error message. Speaking of which, if we don't already, creating aliases (I believe they exist on simplestreams) for corexx to Ubuntu releases (16.04, 18.04, ...).

Let's discuss in 3 weeks

sergiusens (approved):

cmatsuoka (approved):
Thanks Chris, lgtm.

PR #24, opened 2021-04-09
buildd: take sudo secure_path into account for PATH

cjp256:
When using sudo the PATH gets set to a cofigured secure_path, which is
not quite the same as the default found in /etc/environment. Redefine
default_command_environment() to be the combination of the two with the
secure path taking precendence.

By relying on this behavior, we can set /etc/environment and stop
setting the environment out for each command.

-----

### DIFF
--- a/craft_providers/bases/buildd.py
+++ b/craft_providers/bases/buildd.py
@@ -35,12 +35,19 @@ def default_command_environment() -> Dict[str, Optional[str]]:
 
     The minimum environment for the buildd image to be configured and function
     properly.  This contains the default environment found in Ubuntu's
-    /etc/environment.
+    /etc/environment, replaced with the "secure_path" defaults used by sudo for
+    instantiating PATH.  In practice it really just means the PATH set by sudo.
+
+    Default /etc/environment found in supported Ubuntu versions:
+    PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin"
+
+    Default /etc/sudoers found in supported Ubuntu versions:
+    PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
 
     :returns: Dictionary of environment key/values.
     """
     return dict(
-        PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin"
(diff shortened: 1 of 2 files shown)


craft_providers/bases/buildd.py
```diff
@@ -35,12 +35,19 @@ def default_command_environment() -> Dict[str, Optional[str]]:
...
+    PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin"
+
+    Default /etc/sudoers found in supported Ubuntu versions:
+    PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
 
     :returns: Dictionary of environment key/values.
     """
     return dict(
-        PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin"
+        PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin"
```
sergiusens, line 50: I would caveat this with /snap/bin being first on the list (we do so for snapcraft)

sergiusens (approved):

craft_providers/bases/buildd.py
```diff
@@ -35,12 +35,19 @@ def default_command_environment() -> Dict[str, Optional[str]]:
...
     The minimum environment for the buildd image to be configured and function
     properly.  This contains the default environment found in Ubuntu's
-    /etc/environment.
+    /etc/environment, replaced with the "secure_path" defaults used by sudo for
+    instantiating PATH.  In practice it really just means the PATH set by sudo.
+
+    Default /etc/environment found in supported Ubuntu versions:
+    PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin"
+
+    Default /etc/sudoers found in supported Ubuntu versions:
```
cmatsuoka, line 44: Maybe "Default /etc/sudoers secure_path"?

cmatsuoka (approved):
Lgtm, some nitpicking in docstrings.

craft_providers/bases/buildd.py
cjp256: I specifically wanted the default to match the behavior of the base.

Snapcraft or any other application can specify their own custom environment to do whatever makes sense for it.

craft_providers/bases/buildd.py
cjp256: Fixed, thanks!

PR #25, opened 2021-04-09
buildd: introduce os compatibility checks

cjp256:
- Introduce BaseCompatibilityError.
    - Add tests for base errors.

PR #26, opened 2021-04-10
buildd: add instance configuration & perform compatibility check

cjp256:
buildd: add instance configuration & perform compatibility check

A compatibility tag will allow buildd to evolve in the future, providing
an interface to declare a subsequent update as incompatible with the old
one, forcing the user (or application) to clean the instance before
use.  It is possible that this may never increment if compatibility
is forever maintained, but since it cannot be guaranteed, we define this up
front with an extendable InstanceConfiguration for application-specific
customization.

Using class variables to allow application-specific extensions, read the                                           
instance configuration (if available) from /etc/craft-instance.conf using                       
InstanceConfiguration.

Set the compatibility tag to "buildd-base-v0" and provide guidance on how                                     
the applications should extend this tag.

Depends on #25 #24

### DIFF
--- a/craft_providers/bases/buildd.py
+++ b/craft_providers/bases/buildd.py
@@ -21,11 +21,12 @@ import subprocess
 import time
 from textwrap import dedent
 from time import sleep
-from typing import Dict, Optional
+from typing import Dict, Optional, Type
 
 from craft_providers import Base, Executor, errors
 from craft_providers.util.os_release import parse_os_release
 
+from . import instance_config
 from .errors import BaseCompatibilityError, BaseConfigurationError
 
 logger = logging.getLogger(__name__)
@@ -79,10 +80,31 @@ class BuilddBaseAlias(enum.Enum):
 class BuilddBase(Base):
     """Support for Ubuntu minimal buildd images.
 
+    :cvar compatibility_tag: Tag/Version for variant of build configuration and
+        setup.  Any change to this version would indicate that prior [versioned]
+        instances are incompatible and must be cleaned.  As such, any new value
+        should be unique to old values (e.g. incrementing).  It suggested to
+        extend this tag, not overwrite it, e.g.: compatibilty_tag =
+        f"{appname}-{BuildBase.compatibility_tag}.{apprevision}" to ensure base
+        compatibility levels are maintained.
+    :cvar instance_config_path: Path to persistent environment configuration
+        used for compatibility checks (or other data).  Set to
+        /etc/craft-instance.conf, but may be overridden for application-specific
+        reasons.
+    :cvar instance_config_type: Class defining instance configuration.  May be
+        overridden with an application-specific subclass of InstanceConfiguration
+        to enable application-specific extensions.
+
     :param alias: Base alias / version.
     :param environment: Environment to set in /etc/environment.
     :param hostname: Hostname to configure.
     """
+
+    compatibility_tag: str = "buildd-base-v0"
+    instance_config_path: pathlib.Path = pathlib.Path("/etc/craft-instance.conf")
+    instance_config_type: Type[
+        instance_config.InstanceConfiguration
+    ] = instance_config.InstanceConfiguration
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


craft_providers/bases/buildd.py
```diff
@@ -100,6 +122,33 @@ def __init__(
...
+        _check_deadline(deadline)
+        config = self.instance_config_type.load(
+            executor=executor,
+            config_path=self.instance_config_path,
+        )
+
+        # If no config has been written, assume it is compatible (likely an
+        # unfinished setup).
+        if config is None:
+            return
```
cmatsuoka, line 142: Is this safe? What are the consequences of assuming it's compatible when it's not?

craft_providers/bases/buildd.py
```diff
@@ -100,6 +122,33 @@ def __init__(
 
         self.hostname = hostname
 
+    def _ensure_instance_config_compatible(
```
cmatsuoka, line 125: Maybe this could be `_check_instance_config_compatibility`? "Ensure" suggests that this method will make the configuration be compatible if it's not.

craft_providers/bases/buildd.py
```diff
@@ -299,6 +350,19 @@ def _setup_hostname(self, *, executor: Executor, deadline: Optional[float]) -> N
                 details=errors.details_from_called_process_error(error),
             ) from error
 
+    def _setup_instance_config(
+        self, *, executor: Executor, deadline: Optional[float]
+    ) -> None:
+        config = instance_config.InstanceConfiguration(
+            compatibility_tag=self.compatibility_tag
+        )
```
facundobatista, line 358: this multiline thing does fit in 99c, right?

craft_providers/bases/buildd.py
```diff
@@ -299,6 +350,19 @@ def _setup_hostname(self, *, executor: Executor, deadline: Optional[float]) -> N
...
+
+        config.save(
+            executor=executor,
+            config_path=self.instance_config_path,
+        )
```
facundobatista, line 363: same here

facundobatista (approved):
Approving. Please consider NOT breaking the lines if they fit in the 99 columns width.

craft_providers/bases/buildd.py
cjp256: In the case of the instance config, if it's not present, we're dependent on the other checks that are done (i.e. os-release check).

cmatsuoka (approved):
Thanks for the compatibility explanation. Left a couple of comments on optional, cosmetic issues.

craft_providers/bases/buildd.py
sergiusens: I believe this is just black doing its thing

craft_providers/bases/buildd.py
sergiusens: I believe this is just black doing its thing

craft_providers/bases/buildd.py
```diff
@@ -79,11 +80,32 @@ class BuilddBaseAlias(enum.Enum):
...
+        instances are incompatible and must be cleaned.  As such, any new value
+        should be unique to old values (e.g. incrementing).  It suggested to
+        extend this tag, not overwrite it, e.g.: compatibilty_tag =
+        f"{appname}-{BuildBase.compatibility_tag}.{apprevision}" to ensure base
+        compatibility levels are maintained.
+    :cvar instance_config_path: Path to persistent environment configuration
+        used for compatibility checks (or other data).  Set to
+        /etc/craft-instance.conf, but may be overridden for application-specific
+        reasons.
+    :cvar instance_config_type: Class defining instance configuration.  May be
```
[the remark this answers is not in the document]
cjp256: Renamed, thnx

craft_providers/bases/buildd.py
cjp256: I do not have the same interpretation of ensure, but regardless this is a pattern I use fairly regularly. I'll ask the team if there is a preferred approach and update it in the future, if needed.

PR #27, opened 2021-04-12
multipass: doc fix and require check for Multipass._run()

cjp256:
All callers specify check=, but better to require it than
    to change the default.

### DIFF
--- a/craft_providers/multipass/multipass.py
+++ b/craft_providers/multipass/multipass.py
@@ -50,10 +50,10 @@ class Multipass:
         self,
         command: List[str],
         *,
-        check=True,
+        check: bool,
         **kwargs,
     ) -> subprocess.CompletedProcess:
-        """Execute command in instance_name, allowing output to console."""
+        """Execute multipass command."""
         command = [str(self.multipass_path), *command]
 
         logger.debug("Executing on host: %s", shlex.join(command))


craft_providers/multipass/multipass.py
```diff
@@ -50,10 +50,10 @@ def _run(
         self,
         command: List[str],
         *,
-        check=True,
+        check: bool,
         **kwargs,
     ) -> subprocess.CompletedProcess:
-        """Execute command in instance_name, allowing output to console."""
+        """Execute command on host, allowing output to console."""
```
sergiusens, line 56: I don understand, from the point of view of the user of the API, either that or "host" is too generic.

craft_providers/multipass/multipass.py
```diff
@@ -50,10 +50,10 @@ def _run(
         self,
         command: List[str],
         *,
-        check=True,
+        check: bool,
```
sergiusens, line 53: I don't mind this change, but the PR should probably mention why

sergiusens (changes requested):

craft_providers/multipass/multipass.py
cjp256: Agreed, fixed.

craft_providers/multipass/multipass.py
cjp256: Fair enough, simplified to `execute multipass command` :D

sergiusens (approved):

cmatsuoka (approved):

PR #33, opened 2021-04-16
lxd: introduce LXDInstance

cjp256:
- make `lxc` fixture available to all lxd integration tests.

- add unit & integration tests

-----

### DIFF
--- a/craft_providers/lxd/lxd_instance.py
+++ b/craft_providers/lxd/lxd_instance.py
@@ -325,11 +325,8 @@ class LXDInstance(Executor):
     ) -> None:
         """Mount host source directory to target mount point.
 
-        Checks first to see if already mounted. If no device name is
-        given, it will be generated with the format:
-         disk_<target.as_posix().replace("/", "_")>
-
-        For example:  /mnt/foo will be named "disk__mnt_foo"
+        Checks first to see if already mounted.  If no device name is given, it
+        will be generated with the format "disk-{target.as_posix()}".
 
         :param host_source: Host path to mount.
         :param target: Instance path to mount to.
@@ -337,11 +334,11 @@ class LXDInstance(Executor):
 
         :raises LXDError: On unexpected error.
         """
-        if device_name is None:
-            device_name = "disk_" + target.as_posix().replace("/", "_")
-
         if self.is_mounted(host_source=host_source, target=target):
             return
+
+        if device_name is None:
+            device_name = "disk-" + target.as_posix()
 
         self.lxc.config_device_add_disk(
             instance_name=self.name,
--- a/craft_providers/lxd/lxd_instance.py
+++ b/craft_providers/lxd/lxd_instance.py
@@ -288,7 +288,7 @@ class LXDInstance(Executor):
         *,
         image: str,
         image_remote: str,
-        uid: str = str(os.getuid()),
+        map_user_uid: bool = False,
         ephemeral: bool = False,
     ) -> None:
         """Launch instance.
@@ -301,7 +301,10 @@ class LXDInstance(Executor):
         :raises LXDError: On unexpected error.
         """
         config_keys = dict()
-        config_keys["raw.idmap"] = f"both {uid!s} 0"
+
+        if map_user_uid:
+            uid = os.getuid()
+            config_keys["raw.idmap"] = f"both {uid!s} 0"
 
         if self._host_supports_mknod():
             config_keys["security.syscalls.intercept.mknod"] = "true"
--- a/craft_providers/lxd/lxd_instance.py
+++ b/craft_providers/lxd/lxd_instance.py
@@ -413,6 +413,8 @@ class LXDInstance(Executor):
                 f"Directory not found: {str(destination.parent.as_posix())!r}"
             )
 
+        # Copy into target with uid/gid 0, rather than copying the IDs from the
+        # host file.
         self.lxc.file_push(
             instance_name=self.name,
             source=source,
--- a/craft_providers/lxd/lxd_instance.py
+++ b/craft_providers/lxd/lxd_instance.py
@@ -237,14 +237,12 @@ class LXDInstance(Executor):
         """Get state configuration for instance.
 
         :returns: State information parsed from lxc if instance exists, else
-            None.
+                  None.
 
         :raises LXDError: On unexpected error.
         """
         instances = self.lxc.list(project=self.project, remote=self.remote)
 
-        # lxc returns a filter instances starting with instance name rather
-        # than the exact instance.  Find the exact match...
         for instance in instances:
             if instance["name"] == self.name:
                 return instance
@@ -357,7 +355,7 @@ class LXDInstance(Executor):
     def _host_supports_mknod(self) -> bool:
         """Check if host supports mknod in container.
 
-        See: https://actions.linuxcontainers.org/lxd/docs/master/syscall-interception
+        See: https://linuxcontainers.org/lxd/docs/master/syscall-interception
 
         :returns: True if mknod is supported.
 
--- a/craft_providers/lxd/lxd_instance.py
+++ b/craft_providers/lxd/lxd_instance.py
@@ -106,7 +106,6 @@ class LXDInstance(Executor):
         """
         with tempfile.NamedTemporaryFile(delete=False) as temp_file:
             shutil.copyfileobj(content, temp_file)
-            temp_file.flush()
 
         temp_path = pathlib.Path(temp_file.name)
         self.lxc.file_push(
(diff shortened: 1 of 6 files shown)


craft_providers/lxd/lxd_instance.py
```diff
@@ -0,0 +1,491 @@
...
+
+        :returns: State information parsed from lxc if instance exists, else
+            None.
+
+        :raises LXDError: On unexpected error.
+        """
+        instances = self.lxc.list(project=self.project, remote=self.remote)
+
+        # lxc returns a filter instances starting with instance name rather
+        # than the exact instance.  Find the exact match...
```
cmatsuoka, line 248: I understand what you're doing but this comment is confusing, perhaps rephrase it?

craft_providers/lxd/lxd_instance.py
```diff
@@ -0,0 +1,494 @@
...
+
+        :returns: True if instance is running.
+
+        :raises LXDError: On unexpected error.
+        """
+        state = self._get_state()
+        if state is None:
+            raise LXDError(brief=f"Instance {self.name!r} does not exist.")
+
+        return state.get("status") == "Running"
```
cmatsuoka, line 284: Can this be affected by locale?

craft_providers/lxd/lxd_instance.py
```diff
@@ -0,0 +1,494 @@
...
+        target: pathlib.Path,
+        device_name: Optional[str] = None,
+    ) -> None:
+        """Mount host source directory to target mount point.
+
+        Checks first to see if already mounted. If no device name is
+        given, it will be generated with the format:
+         disk_<target.as_posix().replace("/", "_")>
+
+        For example:  /mnt/foo will be named "disk__mnt_foo"
```
cmatsuoka, line 335: Can't this create ambiguities between `/mnt/foo` and a mountpoint called `/mnt_foo`?

craft_providers/lxd/lxd_instance.py
```diff
@@ -0,0 +1,494 @@
...
+            path=target,
+            device=device_name,
+            project=self.project,
+            remote=self.remote,
+        )
+
+    def _host_supports_mknod(self) -> bool:
+        """Check if host supports mknod in container.
+
+        See: https://actions.linuxcontainers.org/lxd/docs/master/syscall-interception
```
cmatsuoka, line 361: Is the `actions` name correct here? It doesn't resolve for me, and the context exists at `linuxcontainers.org`.

craft_providers/lxd/lxd_instance.py
```diff
@@ -0,0 +1,494 @@
...
+
+        :returns: True if mknod is supported.
+
+        :raises LXDError: On unexpected error.
+        """
+        cfg = self.lxc.info(project=self.project, remote=self.remote)
+        env = cfg.get("environment", dict())
+        kernel_features = env.get("kernel_features", dict())
+        seccomp_listener = kernel_features.get("seccomp_listener", "false")
```
cmatsuoka, line 370: The linked documentation doesn't mention `secomp_listener`, shouldn't we be checking `security.syscalls.intercept.mknod` in the instance configuration?

craft_providers/lxd/lxd_instance.py
cjp256: Will do!

craft_providers/lxd/lxd_instance.py
cjp256: Good question, should set it here. will fix!

craft_providers/lxd/lxd_instance.py
cjp256: Good point... need to think on that one a bit.

craft_providers/lxd/lxd_instance.py
cjp256: Will fix

craft_providers/lxd/lxd_instance.py
cjp256: This is to determine if the host kernel has seccomp so we can safely enable mknod in the instance configuration :)

craft_providers/lxd/lxd_instance.py
```diff
@@ -0,0 +1,494 @@
...
+        :param content: Contents of file.
+        :param file_mode: File mode string (e.g. '0644').
+        :param group: File group owner/id.
+        :param user: File user owner/id.
+
+        :raises LXDError: On unexpected error.
+        """
+        with tempfile.NamedTemporaryFile(delete=False) as temp_file:
+            shutil.copyfileobj(content, temp_file)
+            temp_file.flush()
```
facundobatista, line 109: why this flush is necessary? the temp file is closed when the context manager exits, so...

craft_providers/lxd/lxd_instance.py
facundobatista: good catch

facundobatista (changes requested):
Added a couple of details, but the branch is too big :(

craft_providers/lxd/lxd_instance.py
cmatsuoka: Perhaps something similar to what systemd-escape(1) does could help (depending on the set of allowed characters).

craft_providers/lxd/lxd_instance.py
cjp256: I think I prior version of this code used it before exiting.  I can't recall a specific problem so I'm removing it (and the integration tests don't show any issue).

craft_providers/lxd/lxd_instance.py
cjp256: Actually lxc does not seem to have an issue with any characters, including unicode.  So I'm going to remove the replace.

cjp256: Fixed

craft_providers/lxd/lxd_instance.py
cjp256: Fixed

craft_providers/lxd/lxd_instance.py
cjp256: Fixed

craft_providers/lxd/lxd_instance.py
cjp256: Fixed

craft_providers/lxd/lxd_instance.py
cjp256: It does not look like lxc has localization.

craft_providers/lxd/lxd_instance.py
```diff
@@ -0,0 +1,488 @@
...
+        :param group: File group owner/id.
+        :param user: File user owner/id.
+
+        :raises LXDError: On unexpected error.
+        """
+        with tempfile.NamedTemporaryFile(delete=False) as temp_file:
+            shutil.copyfileobj(content, temp_file)
+
+        temp_path = pathlib.Path(temp_file.name)
+        self.lxc.file_push(
```
sergiusens, line 111: is this better than just running "touch" on the destination?

craft_providers/lxd/lxd_instance.py
```diff
@@ -0,0 +1,488 @@
...
+                f"Directory not found: {str(destination.parent.as_posix())!r}"
+            )
+
+        self.lxc.file_push(
+            instance_name=self.name,
+            source=source,
+            destination=destination,
+            project=self.project,
+            remote=self.remote,
+            gid=0,
```
sergiusens, line 422: this should be documented

sergiusens (commented):
added questions/comments

craft_providers/lxd/lxd_instance.py
cjp256: I'm not sure what you mean?  

I just played around with lxd and realized I can use stdin/stdout with file push/pull even though not documented in help.  so I'm going to revisit this in the future separately so we can get rid of the NamedTemporaryFile.

craft_providers/lxd/lxd_instance.py
cjp256: Done.

sergiusens (approved):

craft_providers/lxd/lxd_instance.py
cmatsuoka: Not now, but it may have in the future. Setting `LC_ALL` to something sane such as `C` or `POSIX` before collecting this data is a safer approach, but we can do this later.

cmatsuoka (approved):
Thanks Chris, it looks good. The running status may fail in the future if someone decides to localize it, but we can deal with this case later.

PR #34, opened 2021-04-20
lxd: introduce LXD for interacting with `lxd` command-line

cjp256:
Includes a `is_supported_version()` helper to ensure LXD version
meets minimum required version by craft-providers.

-----

### DIFF
--- a/craft_providers/lxd/lxd.py
+++ b/craft_providers/lxd/lxd.py
@@ -73,8 +73,8 @@ class LXD:
         version_components = version.split(".")
 
         try:
-            major_minor = ".".join([version_components[0], version_components[1]])
-            return float(major_minor) >= 4.0
+            major = int(version_components[0])
+            return major >= 4
         except (ValueError, IndexError) as error:
             raise LXDError(
                 "Failed to parse LXD version.",
--- a/craft_providers/lxd/lxd.py
+++ b/craft_providers/lxd/lxd.py
@@ -1,0 +1,138 @@
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""LXD command-line interface helpers."""
+
+import logging
+import pathlib
+import subprocess
+from typing import Optional
+
+from craft_providers.errors import details_from_called_process_error
+
+from .errors import LXDError
+
+logger = logging.getLogger(__name__)
+
+
+class LXD:
+    """Interface to `lxd` command-line.
+
+    :param lxd_path: Path to lxd.
+    """
+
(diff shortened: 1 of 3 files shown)


craft_providers/lxd/lxd.py
```diff
@@ -0,0 +1,138 @@
...
+            cmd = []
+
+        cmd += [str(self.lxd_path), "init"]
+
+        if auto:
+            cmd.append("--auto")
+
+        try:
+            subprocess.run(cmd, check=True)
+        except subprocess.CalledProcessError as error:
```
sergiusens, line 58: do we have a way to raise a SudoRequired specific error?

tests/integration/lxd/test_lxd.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@pytest.fixture()
+def lxd():
+    yield LXD()
+
+
+def test_init(lxd):
+    lxd.init(auto=True)
+    lxd.init(sudo=True, auto=True)
```
sergiusens, line 28: no assertion to verify it is actually initialized?

sergiusens (approved):
looks good, with comments

tests/integration/lxd/test_lxd.py
cjp256: this is really just checking that init completed without error.  what would you recommend to verify? something like an lxc list?

when I get to the installer tests, there will be a more in-depth test where we install/configure lxd from scratch.

craft_providers/lxd/lxd.py
cjp256: I don't really know of a great way to do that generically because the user may have permissions to do it if they are in the correct group (normally `lxd`).  The best I think could be done is to check `/var/snap/lxd/common/lxd/unix.socket` and see if we have access to open it, but that has a lot of assumptions.  More generically, maybe we could call `lxd waitready --timeout 1` and see if it complains about "permission denied", but that forces us to wait a second and probably has its own issues.

Practically speaking, from an application perspective, the user is granting the application permission to install the provider with sudo when using Linux.  The install() method will assume permission is granted and always use sudo.  We could skip the sudo if running as root, but that logic would probably be better handled there.

cjp256: I put the sudo check into install().

craft_providers/lxd/lxd.py
```diff
@@ -0,0 +1,138 @@
...
+        craft-providers (currently >= 4.0).
+
+        :returns: True if installed version is supported.
+        """
+        version = self.version()
+        version_components = version.split(".")
+
+        try:
+            major_minor = ".".join([version_components[0], version_components[1]])
+            return float(major_minor) >= 4.0
```
cmatsuoka, line 77: For robustness, I'd suggest comparing major and minor separately (for the case where we require the minimum version to be e.g. 4.2 and the current version is 4.10).

cmatsuoka (approved):
Looks good, left a note about comparing version numbers (not a huge fan of using float for that).

craft_providers/lxd/lxd.py
cjp256: Excellent point, will fix

cjp256: For now I limited it to just checking that the major version is >= 4.  If we need to bump the required version beyond 4.0, we add the minor combination check.

sergiusens: `pkg_resources.parse_version` ftw
https://github.com/snapcore/snapcraft/blob/master/snapcraft/internal/elf.py#L121

cjp256: I was thinking about that earlier.  I went ahead and switched to it :D

PR #35, opened 2021-04-20
lxd: introduce launch() interface

cjp256:
Bootstrap with auto_clean, ephemeral, and map_user_uid support.

Future work will incorporate snapshot support.

-----


tests/integration/lxd/test_launcher.py
```diff
@@ -0,0 +1,229 @@
...
+import pytest
+
+from craft_providers import bases, lxd
+
+from . import conftest
+
+
+@pytest.fixture()
+def core20_instance():
+    with conftest.tmp_instance(
```
sergiusens, line 28: why is tmp_instance in conftest?

tests/integration/lxd/test_launcher.py
```diff
@@ -0,0 +1,229 @@
...
+    [
+        pytest.param(
+            bases.BuilddBaseAlias.XENIAL,
+            "16.04",
+        ),
+        (bases.BuilddBaseAlias.BIONIC, "18.04"),
+        (bases.BuilddBaseAlias.FOCAL, "20.04"),
+    ],
+)
+def test_launch(instance_name, alias, image_name):
```
sergiusens, line 53:
```suggestion
def test_launch_and_run(instance_name, alias, image_name):
```

sergiusens (approved):
Looks good, question and a tiny suggestion

tests/integration/lxd/test_launcher.py
cjp256: Because it's shared amongst the fixtures, I figured it was a reasonable spot to put it.  Would you prefer if I moved shared bits into its own module?

craft_providers/lxd/launcher.py
```diff
@@ -0,0 +1,82 @@
...
+        except bases.BaseCompatibilityError as error:
+            if auto_clean:
+                logger.debug(
+                    "Cleaning incompatible container %r (reason: %s).",
+                    instance.name,
+                    error.reason,
+                )
+                instance.delete()
+            else:
+                raise error
```
facundobatista, line 73: To "purely" raise the original error, an empty `raise` is better. Or you're trying to get other behaviour here?

facundobatista (approved):
Approving, but please check inline question. Thanks!

craft_providers/lxd/launcher.py
cjp256: Never really thought about it.  I suppose I always do `raise` when I don't have error and `raise error` when I do, just feels natural.  I have not noticed any unexpected side effects.

`raise` it is :D

cjp256, line 73:
```suggestion
                raise
```

cjp256: Fixed (in two locations).

PR #36, opened 2021-04-20
lxd: introduce installation support

cjp256:

-----


tests/integration/lxd/conftest.py
```diff
@@ -91,13 +93,45 @@ def tmp_instance(
...
+
+    If the environment has CRAFT_PROVIDERS_TESTS_ENABLE_LXD_INSTALL=1,
+    force the installation of LXD if uninstalled.
+    """
+    if sys.platform != "linux":
+        pytest.skip(f"lxd not supported on {sys.platform}")
+
+    if os.environ.get("CRAFT_PROVIDERS_TESTS_ENABLE_LXD_INSTALL") == "1":
+        lxd.install()
+    elif not lxd.is_installed():
```
sergiusens, line 106: would this not silently skip tests, should the conditional be the other way around, skip if the env is not set, falback to just installing

sergiusens (changes requested):
Looks good, just a logic switch comment

tests/integration/lxd/conftest.py
cjp256: If lxd is installed, the tests will run.
If lxd is not installed and CRAFT_PROVIDERS_TESTS_ENABLE_LXD_INSTALL == 1, lxd will be installed and the tests will run.
Otherwise, the tests will be skipped.

There is a bug here I fixed for both Multipass and LXD.

sergiusens (approved):

cmatsuoka (approved):
Thanks Chris, lgtm.

PR #39, opened 2021-04-30
setup: pypi requires valid author email

cjp256:

-----

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,7 @@ setup(
     description="Craft provider tooling",
     long_description=readme,
     author="Canonical Ltd.",
-    author_email="Canonical Ltd.",
+    author_email="snapcraft@lists.snapcraft.io",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -66,7 +66,7 @@
     description="Craft provider tooling",
     long_description=readme,
     author="Canonical Ltd.",
-    author_email="Canonical Ltd.",
+    author_email="chris.patterson@canonical.com",
```
sergiusens, line 69: I would probably use snapcraft@lists.snapcraft.io which is hooked up to the forum so we all get the emails.

cjp256: done

cmatsuoka (approved):
I'll add the same fix to craft-parts when it reaches mvp state.

sergiusens (approved):

PR #41, opened 2021-04-30
pylint: uprev pylint and address new consider-using-with warnings

cjp256:

-----

### DIFF
--- a/craft_providers/multipass/multipass.py
+++ b/craft_providers/multipass/multipass.py
@@ -294,37 +294,37 @@ class Multipass:
         :raises MultipassError: On error.
         """
         command = [str(self.multipass_path), "transfer", source, "-"]
-        proc = subprocess.Popen(
+        with subprocess.Popen(
             command,
             stdin=subprocess.DEVNULL,
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
-        )
+        ) as proc:
 
-        # Should never happen, but mypy/pyright makes noise.
-        assert proc.stdout is not None
-        assert proc.stderr is not None
+            # Should never happen, but mypy/pyright makes noise.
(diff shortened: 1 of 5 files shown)


cjp256:
Depends on #40 for other req updates.

cmatsuoka (approved):

craft_providers/multipass/multipass.py
```diff
@@ -342,33 +342,34 @@ def transfer_source_io(
...
+                    break
+
+                proc.stdin.write(data)
+
+            # Wait until process is complete.
+            try:
+                _, stderr = proc.communicate(timeout=30)
+            except subprocess.TimeoutExpired:
+                proc.kill()
+                _, stderr = proc.communicate()
```
facundobatista, line 364: Do we still need this second `communicate`, taking in consideration that all `fd`s will be closed by the context manager?

craft_providers/multipass/multipass.py
```diff
@@ -342,33 +342,34 @@ def transfer_source_io(
...
+
+            if proc.returncode != 0:
```
facundobatista, line 366: Considering that the have the context manager now to enforce that the process is waited and all crisp and shiny, this block about checking the `returncode` probably should NOT be indented inside the context manager? (just leave it as before, I mean)

facundobatista (approved):
Approved, but please take a look to the questions inline, thanks!

craft_providers/multipass/multipass.py
cjp256: Done.

craft_providers/multipass/multipass.py
cjp256: Good point, fixed!

PR #42, opened 2021-05-03
multipass: is_supported_version() interface and require >=1.7

cjp256:
In preparation for multipass release, add interface for supported version and remove skipped test for multipass issue in now-unsupported versions.

### DIFF
--- a/craft_providers/multipass/multipass.py
+++ b/craft_providers/multipass/multipass.py
@@ -44,7 +44,10 @@ class Multipass:
     """Wrapper for multipass command.
 
     :param multipass_path: Path to multipass command to use.
+    :cvar minimum_required_version: Minimum required version for compatibility.
     """
+
+    minimum_required_version = "1.7"
 
     def __init__(
         self, *, multipass_path: pathlib.Path = pathlib.Path("multipass")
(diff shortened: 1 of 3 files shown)


sergiusens (approved):

sergiusens:
I understand this will fail until the release happens :-)

cjp256:
> I understand this will fail until the release happens :-)

Indeed.  I got excited when I saw the snap refreshed and I thought it happened.  Then realized it was just a rebuild after the test failed :man_facepalming:  Opened the PR as draft just to make sure it is handy when it does land.

sergiusens (approved):

craft_providers/multipass/multipass.py
```diff
@@ -128,6 +133,20 @@ def info(self, *, instance_name: str) -> Dict[str, Any]:
...
+        A helper to check if Multipass meets minimum supported version for
+        craft-providers.
+
+        :returns: True if installed version is supported.
+        """
+        version, _ = self.version()
+
+        return pkg_resources.parse_version(version) >= pkg_resources.parse_version(
+            self.minimum_required_version
+        )
```
cmatsuoka, line 148: Do we have any guarantee that Multipass versions will comply with [PEP 440](https://www.python.org/dev/peps/pep-0440/)? If not, and if it follows semanic versioning, we can use [semantic-version](https://pypi.org/project/semantic-version/) instead.

Looking at past version numbering used in Multipass it seems that it uses plain `major.minor.release` so probably any of these will work. (There was an -rc version but I don't know if that appears in the actual version string.)

cmatsuoka (approved):

PR #44, opened 2021-05-07
integration tests: consolidate lxd/multipass fixtures for reuse

cjp256:

-----

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -12,7 +12,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Fixtures for LXD integration tests."""
+"""Fixtures for integration tests."""
 import os
 import pathlib
 import random
@@ -27,12 +27,17 @@ from craft_providers import lxd, multipass
 
 
 def generate_instance_name():
+    """Generate a random instance name."""
     return "itest-" + "".join(random.choices(string.ascii_uppercase, k=8))
 
 
 @pytest.fixture()
 def home_tmp_path():
-    """Multipass doesn't have access to /tmp."""
+    """Provide a temporary directory located in user's home directory.
+
+    Multipass doesn't have access to /tmp, this fixture provides an equivalent
+    to tmp_path for tests that require it.
+    """
     with tempfile.TemporaryDirectory(
         suffix=".tmp-pytest", dir=pathlib.Path.home()
     ) as temp_dir:
@@ -41,11 +46,13 @@ def home_tmp_path():
 
 @pytest.fixture()
 def instance_name():
+    """Provide a random name for an instance to launch."""
     yield generate_instance_name()
 
 
 @pytest.fixture(scope="module")
 def reusable_instance_name():
+    """Provide a random name for an instance to launch with scope=module."""
     yield generate_instance_name()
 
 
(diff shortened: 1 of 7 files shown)


cjp256:
Also includes a doc fix.

tests/integration/conftest.py
```diff
@@ -0,0 +1,137 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Fixtures for LXD integration tests."""
```
sergiusens, line 15: description update perhaps?

tests/integration/conftest.py
```diff
@@ -0,0 +1,137 @@
...
+from craft_providers import lxd, multipass
+
+
+def generate_instance_name():
+    return "itest-" + "".join(random.choices(string.ascii_uppercase, k=8))
+
+
+@pytest.fixture()
+def home_tmp_path():
+    """Multipass doesn't have access to /tmp."""
```
sergiusens, line 35: this is more of a code comment than a docstring for the test, right?

cjp256: Fixed!

tests/integration/conftest.py
cjp256: Fixed!

cmatsuoka (approved):
Looks good!

tests/integration/conftest.py
```diff
@@ -0,0 +1,144 @@
...
+    """Uninstall Lxd prior to test, if environment allows it.
+
+    Environment may enable this fixture with:
+    CRAFT_PROVIDERS_TESTS_ENABLE_LXD_UNINSTALL=1
+    """
+    if sys.platform != "linux":
+        pytest.skip(f"lxd not supported on {sys.platform}")
+
+    if not lxd.is_installed():
+        pytest.skip("lxd not installed, skipped")
```
facundobatista, line 89: if this fixture provides "an enviroment without LXD", shouldn't be a noop if LXD is just not installed, instead of making the test to skip?

tests/integration/conftest.py
```diff
@@ -0,0 +1,144 @@
...
+
+@pytest.fixture
+def uninstalled_multipass():
+    """Uninstall Multipass prior to test, if environment allows it.
+
+    Environment may enable this fixture with:
+    CRAFT_PROVIDERS_TESTS_ENABLE_MULTIPASS_UNINSTALL=1
+    """
+    if not multipass.is_installed():
+        pytest.skip("multipass not installed, skipped")
```
facundobatista, line 128: Same than above.

facundobatista (commented):
I added a question about the "uninstalled*" fixtures, thanks!

tests/integration/conftest.py
cjp256: > if this fixture provides "an enviroment without LXD", shouldn't be a noop if LXD is just not installed, instead of making the test to skip?

It's for the same reason the lxd tests only run if lxd is installed.  If lxd is installed when tests are invoked, we assume lxd can run OK.

facundobatista: Ok. Maybe worths a sentence in the function's docstring, something like "in case of LXD not being already installed, we assume that it cannot be installed and skip the test"

facundobatista (approved):

tests/integration/conftest.py
cjp256: Done!

tests/integration/conftest.py
cjp256: Done!

PR #45, opened 2021-05-14
actions: introduce helpers to install & inject snaps (CRAFT-253)

cjp256:
- add snap_installer.install_from_store() to install a snap in a
  target environment, pulling from the store.

- add snap_installer.inject_from_host() to install snap from the
  currently-installed version on the host.  First attempts to
  download the snap using the snapd API, falling back to `snap pack`
  in case the snapd API fails (as happens when a snap is installed
  using `snap try`.

- add core20_lxd_instance fixture for integration tests to provide
  integration test environment for helpers.

- add installed_snap contextmanager fixture to install snaps on host
  for a guest.

- add integration & unit test coverage.

-----


craft_providers/actions/snap_installer.py
```diff
@@ -0,0 +1,202 @@
...
+                f"--channel={channel}",
+                f"--basename={snap_name}",
+                "--target-directory=/tmp",
+            ],
+            check=True,
+            capture_output=True,
+        )
+
+        install_cmd = ["snap", "install", f"/tmp/{snap_name}.snap", "--dangerous"]
+        if classic:
```
cmatsuoka, line 190: We always use `--classic` when injecting and it's ignored when the snap is strictly confined, why can't we do the same here?

cmatsuoka (approved):
It looks good, added a couple of questions but I didn't see any showstopper here.

craft_providers/actions/snap_installer.py
cjp256: Ah, good point.  Didn't realize that :D

craft_providers/actions/snap_installer.py
```diff
@@ -0,0 +1,202 @@
...
+    try:
+        # Clean outdated snap, if exists.
+        executor.execute_run(
+            ["rm", "-f", target_snap_path],
+            check=True,
+            capture_output=True,
+        )
+
+        with _get_host_snap(snap_name) as host_snap_path:
+            executor.push_file(
```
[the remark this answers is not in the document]
cjp256: Agreed, :+1: can do separately

craft_providers/actions/snap_installer.py
cjp256: Fixed

sergiusens: it used be an error

cjp256: Yeah I was surprised it didn't cause an issue with the integration tests

craft_providers/actions/snap_installer.py
```diff
@@ -0,0 +1,191 @@
...
+
+def inject_from_host(*, executor: Executor, snap_name: str, classic: bool) -> None:
+    """Inject snap from host snap.
+
+    :raises SnapInstallationError: on unexpected error.
+    """
+    target_snap_path = pathlib.Path(f"/tmp/{snap_name}.snap")
+
+    try:
+        # Clean outdated snap, if exists.
```
sergiusens, line 113: add a TODO here if doing separately

craft_providers/actions/snap_installer.py
```diff
@@ -0,0 +1,191 @@
...
+            ),
+            check=True,
+            capture_output=True,
+        )
+    except subprocess.CalledProcessError as error:
+        raise SnapInstallationError(
+            brief=f"Failed to inject snap {snap_name!r}.",
+            details=details_from_called_process_error(error),
+        ) from error
+    except ProviderError as error:
```
sergiusens, line 138: Remind me what can raise a ProviderError here? Would it not be relevant to add "details"?

craft_providers/actions/snap_installer.py
```diff
@@ -0,0 +1,191 @@
...
+    :param snap_name: Name of snap to install.
+    :param channel: Channel to install from.
+    :param classic: Install in classic mode.
+
+    :raises SnapInstallationError: on unexpected error.
+    """
+    target_snap_path = pathlib.Path(f"/tmp/{snap_name}.snap")
+
+    try:
+        executor.execute_run(
```
sergiusens, line 165: I believe "snap download" caches and knows how to resume, why not download on host and push to avoid downloading the same thing potentially N times? Auto prunning should be fine.

requirements-dev.txt
```diff
@@ -1,23 +1,23 @@
 alabaster==0.7.12
 appdirs==1.4.4
 astroid==2.5.6
-attrs==21.1.0
+attrs==21.2.0
```
sergiusens, line 4: I believe these are left overs.

tests/integration/actions/test_snap_installer.py
```diff
@@ -0,0 +1,142 @@
...
+    snap_name = "craft-integration-test-snap"
+
+    if shutil.which("snap") is None or sys.platform != "linux":
+        pytest.skip("requires linux and snapd")
+
+    if not os.environ.get("CRAFT_PROVIDERS_TESTS_ENABLE_SNAP_INSTALL") == "1":
+        pytest.skip("need permission to install snaps, skipped")
+
+    if pathlib.Path("/snap", snap_name).exists():
+        pytest.skip(f"test snap {snap_name} already exists, please remove")
```
sergiusens, line 43: You could potentially make this a fixture and reduce the copy/paste of this structure. Different PR should be fine

craft_providers/actions/snap_installer.py
cjp256: It's what the `rm` command below is doing.

craft_providers/actions/snap_installer.py
cjp256: push_file() will raise a ProviderError for whatever error may be raised during the file copy.  It shouldn't happen normally, but it would contain the reason (as perhaps explained by multipass/lxd command output).   I chain it so the data is available if necessary.

It would be clearer if I didn't combine the two actions, so I'll do that and add a detail there.

craft_providers/actions/snap_installer.py
cjp256: I was looking ahead to cross-platform support where the host would not support snaps.

Generally, if the host supports snaps, the application is likely to be injected rather than installed from store.

requirements-dev.txt
cjp256: Indeed, resolved conflicts.

tests/integration/actions/test_snap_installer.py
cjp256: Will do.

craft_providers/actions/snap_installer.py
cjp256: Done & added a test.

requirements-dev.txt
cjp256: Done.

craft_providers/actions/snap_installer.py
sergiusens: snap download should work on osx and windows.

tests/integration/actions/test_snap_installer.py
sergiusens: but where?

craft_providers/actions/snap_installer.py
cjp256: Alright, if there is no issue with making snap a dependency, I update to use the host np

tests/integration/actions/test_snap_installer.py
cjp256: I'm thinking I can do this by updating the installed_snap() fixture to take an optional `try_dir` kwarg to install a snap from a directory using try.

cjp256: Done.

sergiusens (approved):

PR #47, opened 2021-05-17
util: add temp path helpers for creating temp paths in home directory

cjp256:
A few places need temp files/dirs and create them in the home directory
for Multipass to be able to access.

This adds two helpers to do this consistently:
- home_temporary_file()
- home_temporary_directory()

-----

### DIFF
--- a/craft_providers/util/temp_paths.py
+++ b/craft_providers/util/temp_paths.py
@@ -28,13 +28,3 @@ def home_temporary_directory() -> Iterator[pathlib.Path]:
         dir=pathlib.Path.home(),
     ) as tmp_dir:
         yield pathlib.Path(tmp_dir)
-
-
-@contextlib.contextmanager
-def home_temporary_file(name: str) -> Iterator[pathlib.Path]:
-    """Create temporary directory in home directory where Multipass has access.
-
-    :param name: Name of file to use (in temporary directory).
-    """
-    with home_temporary_directory() as tmp_dir:
-        yield tmp_dir / name
--- a/craft_providers/util/temp_paths.py
+++ b/craft_providers/util/temp_paths.py
@@ -1,0 +1,40 @@
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


craft_providers/util/temp_paths.py
```diff
@@ -0,0 +1,40 @@
...
+    with tempfile.TemporaryDirectory(
+        suffix=".tmp-craft",
+        dir=pathlib.Path.home(),
+    ) as tmp_dir:
+        yield pathlib.Path(tmp_dir)
+
+
+@contextlib.contextmanager
+def home_temporary_file(name: str) -> Iterator[pathlib.Path]:
+    """Create temporary directory in home directory where Multipass has access.
```
cmatsuoka, line 35: I believe this should be "path in temporary directory"? We're not creating the temporary file and only returning its path, perhaps we should follow what `tempfile.NamedTemporaryFile` does and return a file object instead. If we only get a path back the caller could easily append it using the `home_temporary_directory` context manager directly.

craft_providers/util/temp_paths.py
```diff
@@ -0,0 +1,40 @@
...
+
+    :param name: Name of file to use (in temporary directory).
+    """
+    with home_temporary_directory() as tmp_dir:
+        yield tmp_dir / name
```
cmatsuoka, line 40: We can also verify if the file already exists, but if we only return a path instead of a file we may  incur a TOCTOU race.

cjp256: Ah, well that shouldn't be a problem because the temp dir is guaranteed, and the caller is the sole user.

But based on your earlier comment, maybe it's worth just nuking this helper and stick with home_temporary_directory

craft_providers/util/temp_paths.py
cjp256: Ah, well that shouldn't be a problem because the temp dir is guaranteed (it is created) and the caller is the sole user.

But maybe this helper isn't even worth having, and just stick with `home_temporary_directory()`.

cjp256: Removed for now.

craft_providers/util/temp_paths.py
cjp256: Removed for now.

cmatsuoka (approved):
Thanks!

facundobatista (approved):
Good!

PR #51, opened 2021-05-28
lxc: add has_image() helper and use yaml BaseLoader (CRAFT-144)

cjp256:
The image list data is a little complex for a caller to simply
check if an image is available.  This adds has_image() helper
to do that check.

While parsing image data, LXD tends to trip up Yaml's Loader
when parsing timestamps into datetime format. Use the BaseLoader
instead to avoid parsing timestamps (which may be invalid) and
simply keep them as strings.

-----


facundobatista (changes requested):
Code looks fine, but we should add test cases for the "broken timestamp", so we're sure we won't come back to regular yaml parsing in the future.

cjp256:
> Code looks fine, but we should add test cases for the "broken timestamp", so we're sure we won't come back to regular yaml parsing in the future.

Done!

cmatsuoka (approved):
Looks good!

facundobatista (approved):
Thanks!

PR #52, opened 2021-05-29
lxd: add snapshot support (CRAFT-144)

cjp256:

-----

### DIFF
--- a/craft_providers/lxd/launcher.py
+++ b/craft_providers/lxd/launcher.py
@@ -18,9 +18,66 @@ import logging
 
 from craft_providers import Base, bases
 
+from .lxc import LXC
 from .lxd_instance import LXDInstance
 
 logger = logging.getLogger(__name__)
+
+
+def _formulate_snapshot_image_name(
+    *, image_name: str, image_remote: str, compatibility_tag: str
+) -> str:
+    """Compute snapshot image's name.
+
+    It must take into account each of the following params list below.  Note
+    that the image_name should incorporate the architecture to ensure uniqueness
+    in case more than one arch is supported on the platform (e.g. LXD cluster).
+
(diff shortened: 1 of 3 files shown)


craft_providers/lxd/launcher.py
```diff
@@ -32,6 +89,8 @@ def launch(
     auto_clean: bool = False,
     ephemeral: bool = False,
     map_user_uid: bool = False,
+    use_snapshots: bool = False,
+    lxc: LXC = LXC(),
```
facundobatista, line 93: Please document also this parameter below, thanks!

facundobatista (approved):
Approved, but please see comment, thanks!

sergiusens (approved):
Would be nice to see a ref counter to garbage collect if it is not the latest snapshot :-)

craft_providers/lxd/launcher.py
cjp256: Ah yes, will do thnx

cjp256:
> Would be nice to see a ref counter to garbage collect if it is not the latest snapshot :-)

I don't think there is a need to persist a snapshot for any reason, so the fact that it was used to instantiate an existing build container doesn't really need to factor into the story about when to delete it, though it could.    If the clean story uses a reference counter, the primary drawback imo is that it would limit the utility of the snapshots if I'm only building one project and regularly do a `clean` or something like the `--clean-on-exit`.

I've got two thoughts for handling this:

(1) provide a toggle to delete all existing snapshots when creating a new one if they exist (this could be determined by wildcarding the compatibility tag). Something like `auto_clean_snapshots=True`

(2) provide a more convenient interface for the application to specify a list of snapshots considered valid, and delete the rest (or provide a set no longer valid).

I'm thinking (1) would be the most useful long-term option unless intending on supporting multiple compatibility tags simultaneously.

PR #53, opened 2021-06-17
buildd: install udev dependency for snapd (CRAFT-288)

cjp256:
sudo and udev packages are required for LXD buildd images.
Subsequent PRs will include integration test coverage.

-----

### DIFF
--- a/craft_providers/bases/buildd.py
+++ b/craft_providers/bases/buildd.py
@@ -467,6 +467,7 @@ class BuilddBase(Base):
                     "install",
                     "-y",
                     "fuse",
+                    "udev",
                 ],
                 check=True,
                 capture_output=True,
--- a/tests/unit/bases/test_buildd.py
+++ b/tests/unit/bases/test_buildd.py
@@ -115,7 +115,7 @@ def test_setup(  # pylint: disable=too-many-arguments
         [*DEFAULT_FAKE_CMD, "apt-get", "install", "-y", "apt-utils"]
     )
     fake_process.register_subprocess(
-        [*DEFAULT_FAKE_CMD, "apt-get", "install", "-y", "fuse"]
+        [*DEFAULT_FAKE_CMD, "apt-get", "install", "-y", "fuse", "udev"]
     )
     fake_process.register_subprocess(
         [*DEFAULT_FAKE_CMD, "systemctl", "enable", "systemd-udevd"]
@@ -446,7 +446,7 @@ def test_setup_snapd_failures(
     return_codes[fail_index] = 1
 
     fake_process.register_subprocess(
-        [*DEFAULT_FAKE_CMD, "apt-get", "install", "-y", "fuse"],
+        [*DEFAULT_FAKE_CMD, "apt-get", "install", "-y", "fuse", "udev"],
         returncode=return_codes[0],
     )
     fake_process.register_subprocess(


facundobatista (approved):
+1

tests/unit/bases/test_buildd.py
```diff
@@ -115,7 +115,10 @@ def test_setup(  # pylint: disable=too-many-arguments
         [*DEFAULT_FAKE_CMD, "apt-get", "install", "-y", "apt-utils"]
     )
     fake_process.register_subprocess(
-        [*DEFAULT_FAKE_CMD, "apt-get", "install", "-y", "fuse"]
+        [*DEFAULT_FAKE_CMD, "apt-get", "install", "-y", "sudo"]
+    )
+    fake_process.register_subprocess(
+        [*DEFAULT_FAKE_CMD, "apt-get", "install", "-y", "fuse", "udev"]
```
sergiusens, line 121: when I asked a long time ago you told me udev was no longer needed, why is the reason to add it back?

craft_providers/bases/buildd.py
```diff
@@ -521,6 +523,25 @@ def _setup_snapd(
...
+    def _setup_sudo(self, *, executor: Executor, deadline: Optional[float]) -> None:
+        """Install & configure sudo.
+
+        :param executor: Executor for target container.
+        :param deadline: Optional time.time() deadline.
+        """
+        _check_deadline(deadline)
+        try:
+            executor.execute_run(
+                ["apt-get", "install", "-y", "sudo"],
```
sergiusens, line 535: why do we need sudo? we are already root.

sergiusens (changes requested):
I would like some clarifications in there changes

tests/unit/bases/test_buildd.py
cjp256: My understanding was that you questioned if it was still needed.  I removed it, tests worked so I thought OK, then removed it.  What I failed to realize was that I had lost the the buildd integration tests along the way and it wasn't part of the test coverage like I thought it was... :( I've got another PR to fix another regression wrt buildd in my queue and yet another to fix the coverage gap.

craft_providers/bases/buildd.py
cjp256: We are using sudo to standardize the behavior of exec'ing commands across the providers. Namely, the LXD-provided environment is quite a bit different than running `sudo cmd` for Multipass.  By using sudo for both, we get comparable environments.

cjp256: Removed sudo!

sergiusens (approved):
An integration test has failed. Might be worth looking into

cjp256:
Odd, looks like the ephemeral instance was stopped, but not yet "deleted".  Then of course the cleanup call to delete fails...

```
E           craft_providers.lxd.errors.LXDError: Failed to delete instance 'itest-SEHQQBAD'.
E           * Command that failed: 'lxc --project default delete local:itest-SEHQQBAD --force'
E           * Command exit code: 1
E           * Command standard error output: b'Error: Query deleted 0 rows instead of 1\n'E           craft_providers.lxd.errors.LXDError: Failed to delete instance 'itest-SEHQQBAD'.
E           * Command that failed: 'lxc --project default delete local:itest-SEHQQBAD --force'
E           * Command exit code: 1
E           * Command standard error output: b'Error: Query deleted 0 rows instead of 1\n'
```

First time that's happened AFAIK. I'll restart the tests and see if it repros.

PR #54, opened 2021-06-18
lxd instance: drop sudo usage (CRAFT-289)

cjp256:
This will remove the sudo dependency for LXD, which is not in
the LXD buildd images.

As using sudo populates the processes with /etc/environment,
a subsequent PR will will improve environment handling to
ensure consistency across providers.

Signed-off-by: Chris Patterson <chris.patterson@canonical.com>

- [ ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

-----

PR #55, opened 2021-06-18
lxd: add buildd remote support (CRAFT-260)

cjp256:
Add configure_buildd_image_remote() helper to add the buildd
remote and return its name.  Calling applications will call
this as part of their setup code prior to launching
an instance.

Add launcher integration tests for the 16.04 and 18.04 buildd
images.  Xenial is currently unsupported because its
/etc/apt/sources.list does not include the -updates
or -security suites, breaking an assumption that snapd is updated
in the buildd image (it ships 2.0.2).  Further work will be
required to support it properly.

-----

Requires #54 and #53 for integration tests to pass.

### DIFF
--- a/craft_providers/lxd/__init__.py
+++ b/craft_providers/lxd/__init__.py
@@ -20,3 +20,4 @@ from .launcher import launch  # noqa: F401
 from .lxc import LXC  # noqa: F401
 from .lxd import LXD  # noqa: F401
 from .lxd_instance import LXDInstance  # noqa: F401
+from .remotes import configure_buildd_image_remote  # noqa: F401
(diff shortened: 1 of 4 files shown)


facundobatista (approved):
Looks great!

sergiusens (approved):
I do question the need for the remote, but this looks good, I would consider renaming the remote to something like "craft" so a user can quickly identify why it is there.

cjp256:
> I do question the need for the remote, but this looks good, I would consider renaming the remote to something like "craft" so a user can quickly identify why it is there.

Renamed it to `craft-com.ubuntu.cloud-buildd`

sergiusens:
that works, one test did not :-)

cmatsuoka (approved):

PR #61, opened 2021-06-23
lxd instance: account for command environment (CRAFT-309)

cjp256:
When exec'ing a command through LXD, /etc/environment is not read in
by default as it was when previously using sudo. For now, extend
LXDInstance with the default command environment, which can be
overriden by the execute functions' env= parameter.

- Extend Base definition with 'environment' parameter already present
  in BuilddBase.

- Move _rootify_lxc_command() into LXDInstance and account for the
  default command environment.

- Update LXD launcher to set LXDInstance's environment from the
  existing base configuration.  This way the environment will be
  set for all executed commands.

Future work may change the mechanism used to set the environment
(such as using lxc configuration, or wrapper script).

-----

### DIFF
--- a/craft_providers/base.py
+++ b/craft_providers/base.py
@@ -19,7 +19,7 @@
 
 import logging
 from abc import ABC, abstractmethod
-from typing import Optional
+from typing import Dict, Optional
 
 from .executor import Executor
 
@@ -49,9 +49,18 @@ class Base(ABC):
         extend this tag, not overwrite it, e.g.: compatibilty_tag =
         f"{appname}-{Base.compatibility_tag}.{apprevision}" to ensure base
         compatibility levels are maintained.
+
+    :param environment: Command environment.
     """
 
     compatibility_tag: str = "base-v0"
+
+    def __init__(
+        self,
+        *,
+        environment: Dict[str, Optional[str]],
+    ):
+        self.environment = environment
 
     @abstractmethod
     def setup(
--- a/craft_providers/base.py
+++ b/craft_providers/base.py
@@ -49,18 +49,19 @@ class Base(ABC):
         extend this tag, not overwrite it, e.g.: compatibilty_tag =
         f"{appname}-{Base.compatibility_tag}.{apprevision}" to ensure base
         compatibility levels are maintained.
-
-    :param environment: Command environment.
     """
 
     compatibility_tag: str = "base-v0"
 
-    def __init__(
+    @abstractmethod
+    def get_command_environment(
         self,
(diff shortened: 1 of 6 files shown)


craft_providers/base.py
```diff
@@ -49,10 +49,19 @@ class Base(ABC):
...
+
+    :param environment: Command environment.
     """
 
     compatibility_tag: str = "base-v0"
 
+    def __init__(
+        self,
+        *,
+        environment: Dict[str, Optional[str]],
```
facundobatista, line 61: It's strange that this parameter is mandatory here, and optional in the inherited one. 

What would be the behaviours of other classes inheriting this? Because maybe we could make it optional everywhere, and *here* get the `default_command_environment()` if none was passed.

cjp256: I considered that, but default_command_environment() in theory is specific BuilddBase.  If it were say a Windows instance (e.g. `WindowsBase`), that implementation would define its own default_command_environment.  The common denominator is that the base `environment` should always be defined for a base, but each base may carry its own default (allowing it to be optional in the subclass).

craft_providers/lxd/lxd_instance.py
```diff
@@ -86,6 +64,42 @@ def __init__(
...
+        """Wrap a command to run as root with specified environment.
+
+        Formulate command to:
+
+            - Use sudo to run as root (LXD defaults to root, but doesn't
+              initialize environment like sudo will, e.g. read
+              /etc/environment).  For consistency with Multipass and other
+              providers, always use sudo.
+
+            - Configure sudo to set home directory.
```
cmatsuoka, line 82: I'm a bit confused by this comment, I don't see this happening in the implementation. Is it being done elsewhere?

cjp256: Ah, that's outdated, will fix.  Up until recently it used sudo.

cjp256: Updated the docs and renamed the method too.

cmatsuoka (approved):
Thanks, looks good!

craft_providers/base.py
sergiusens: from an API point of view I would instead do

```python
@abc.abstractmethod
def get_environment(self) -> Dict[str, Optional[str]]
    """Return the command environment."""
```

to require its implementation in the classes that inherit it

cjp256: Fair point, done!

sergiusens (approved):

PR #66, opened 2021-07-09
buildd base: configure apt to error on fetch failures (CRAFT-344)

cjp256:
apt update will return exit code 0 if it fails to fetch updates due to
networking issues.  This is problematic because it can silently fail
during use if networking is not available when running update.

Use the option APT::Update::Error-Mode="any" to catch any fetch failures
and exit with error.

-----


facundobatista (approved):
Nice!

sergiusens (approved):

PR #69, opened 2021-07-28
lxd: ensure user is permitted to use lxd (CRAFT-374)     

cjp256:
- Introduce is_user_permitted() to check if user has the necessary
      permissions (that is, either root or in the lxd group).
    
    - As the number of checks are increasing, add another interface
      ensure_lxd_is_ready() which craft apps can use to verify the readiness
      of LXD prior to use.
    
    - Add permission check to LXD install to ensure user has necessary
      permissions before continuing.
    
    - Remove the initialization check from the launcher in favor of
      using ensure_lxd_is_ready() to catch potential errors even earlier.

### DIFF
--- a/craft_providers/lxd/installer.py
+++ b/craft_providers/lxd/installer.py
@@ -29,6 +29,10 @@ from . import errors
 from .lxc import LXC
 from .lxd import LXD
 
+# grp does not exist on Windows.
+if sys.platform == "linux":
+    import grp
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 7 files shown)


craft_providers/lxd/installer.py
```diff
@@ -89,3 +98,41 @@ def is_installed() -> bool:
...
+def is_user_permitted() -> bool:
+    """Check if user has permisisons to connect to LXD.
+
+    The user must be root or be in lxd group to talk to the local LXD socket.
+
+    :returns: True if user has correct permissions.
+    """
+    return (
+        any(grp.getgrgid(g).gr_name == "lxd" for g in os.getgroups())
+        or os.geteuid() == 0
```
cmatsuoka, line 112: Probably better to test for euid first so you don't have to check groups if user is root?

cmatsuoka (approved):

craft_providers/lxd/installer.py
cjp256: fair point!

cjp256: done!

sergiusens (approved):

PR #70, opened 2021-08-02
Bump version: 1.0.1 → 1.0.2

cjp256:

-----

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.0.1
+current_version = 1.0.2
 commit = True
 tag = True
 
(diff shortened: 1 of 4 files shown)


sergiusens (approved):

facundobatista (approved):
+1, but it concerns me that we need to change this value in four different places. Are we sure they are not five and we're forgetting one? My point is: can we have a single source of truth here? Thanks!

cjp256:
> +1, but it concerns me that we need to change this value in four different places. Are we sure they are not five and we're forgetting one? My point is: can we have a single source of truth here? Thanks!

We're using bumpversion to automate this.
https://github.com/canonical/craft-providers/blob/main/.bumpversion.cfg

PR #73, opened 2021-08-06
util/env_cmd: add support for --chdir (CRAFT-449)

cjp256:
Extend env command wrapper with option for --chdir.

It is the caller's responsibility to ensure that the env command
supports chdir as older versions do not support it (Ubuntu 16.04).

-----

### DIFF
--- a/craft_providers/util/env_cmd.py
+++ b/craft_providers/util/env_cmd.py
@@ -16,11 +16,15 @@
 #
 
 """Helper(s) for env command."""
+import pathlib
 from typing import Dict, List, Optional
 
 
 def formulate_command(
-    env: Optional[Dict[str, Optional[str]]] = None, *, ignore_environment: bool = False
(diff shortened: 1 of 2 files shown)


cmatsuoka (approved):
Code looks good, however this may cause problems in e.g. macos since [its `env` command](https://ss64.com/osx/env.html) doesn't support the feature. If it's used to implement essential features maybe a different solution should be adopted.

cjp256:
> Code looks good, however this may cause problems in e.g. macos since [its `env` command](https://ss64.com/osx/env.html) doesn't support the feature. If it's used to implement essential features maybe a different solution should be adopted.

Agreed.  Should we need this functionality in the future of cross-OS compat we would need to get a proper interface for Multipass to support it like LXD.  I've gotta compile a wishlist :)

sergiusens (approved):

PR #74, opened 2021-08-09
multipass instance: use env's chdir for exec cwd (CRAFT-450)

cjp256:
To support the use of `cwd` keyword arg for execute functions,
use env_cmd's chdir functionality.  This does not support all
OSs (works on 18.04+), but since it will otherwise fail
unexpectedly, this is a reasonable measure for the interim.

Applications needing support for older OSes can use their own
custom launcher script or avoid `cwd` altogether.

A better long term solution is to ask Multipass for additional
interfaces like --cwd <dir> (and --user <user> even) to match
what is available in LXD.

-----

### DIFF
--- a/craft_providers/multipass/multipass_instance.py
+++ b/craft_providers/multipass/multipass_instance.py
@@ -20,7 +20,7 @@ import io
 import logging
 import pathlib
 import subprocess
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional
 
 from craft_providers import errors
 from craft_providers.util import env_cmd
@@ -35,7 +35,7 @@ logger = logging.getLogger(__name__)
 def _rootify_multipass_command(
     command: List[str],
     *,
-    cwd: Union[pathlib.Path, str, None] = None,
+    cwd: Optional[pathlib.Path] = None,
     env: Optional[Dict[str, Optional[str]]] = None,
 ) -> List[str]:
     """Wrap a command to run as root with specified environment.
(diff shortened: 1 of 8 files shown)


craft_providers/multipass/multipass_instance.py
```diff
@@ -189,9 +195,11 @@ def execute_run(
         :raises subprocess.CalledProcessError: if command fails and check is
             True.
         """
+        cwd = kwargs.pop("cwd", None)
```
facundobatista, line 198: Why `cwd` can't be expressed in the method's signature?

craft_providers/multipass/multipass_instance.py
```diff
@@ -35,6 +35,7 @@
 def _rootify_multipass_command(
     command: List[str],
     *,
+    cwd: Union[pathlib.Path, str, None] = None,
```
facundobatista, line 38: Shouldn't these two be ... ?
```
    cwd: Optional[Union[pathlib.Path, str]] = None,
    env: Optional[Dict[str, Union[str, None]]] = None,
```

craft_providers/multipass/multipass_instance.py
```diff
@@ -50,8 +51,13 @@ def _rootify_multipass_command(
     """
     sudo_cmd = ["sudo", "-H", "--"]
 
-    if env is not None:
-        sudo_cmd += env_cmd.formulate_command(env)
+    if cwd is not None:
+        cwd_path: Optional[pathlib.Path] = pathlib.Path(cwd)
```
facundobatista, line 55: I don't understand the usage of `Optional` here...

facundobatista (changes requested):
Added some questions, thanks!

cmatsuoka (approved):
Implementation is fine, but the approach may need to be reworked when/if macos support is required. (Probably the best solution would be to add missing functionality to Multipass.)

craft_providers/multipass/multipass_instance.py
cjp256: Good point, simplifies some things with a bit of extra work. done! :D

craft_providers/multipass/multipass_instance.py
cjp256: They're the same.

craft_providers/multipass/multipass_instance.py
cjp256: reworked anyways to avoid this

facundobatista (approved):
Looks great!

PR #75, opened 2021-08-09
multipass: introduce ensure_multipass_is_ready() (CRAFT-438)

cjp256:
Add checks to ensure Multipass is ready for use similar to LXD's.

-----

### DIFF
--- a/craft_providers/multipass/_ready.py
+++ b/craft_providers/multipass/_ready.py
@@ -1,0 +1,48 @@
+#
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation,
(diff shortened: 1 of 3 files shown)


craft_providers/multipass/_ready.py
```diff
@@ -0,0 +1,48 @@
...
+import logging
+
+from . import errors
+from .installer import is_installed
+from .multipass import Multipass
+
+logger = logging.getLogger(__name__)
+
+
+def ensure_multipass_is_ready(*, multipass: Multipass = Multipass()) -> None:
```
sergiusens, line 29: why is this not an instance method in Multipass?

cjp256: This was to match the approach used with LXD, it certainly could (should?) be after considering it further.

cjp256: This change would result in cross coupling between installer module & Multipass

sergiusens, line 29: I still wonder why we have a default here and if you could just require it. I would argue that being the of the incorrect version is not not being ready and maybe a name change to `ensure_multipass_environment` or similar would be a bit more appropriate.

sergiusens (approved):

facundobatista (approved):
I like it. 

I see the function as "ensure multipass is ready for me to be used" (which includes being a supported version) and not "ready" as in "up and running"...

craft_providers/multipass/_ready.py
cjp256: I agree the name is a bit odd.  I'd like to wrap up this functionality into a generalized `provider.initialize()` API, but I have been delaying that until things have firmed up on the Charmcraft side and we expand to other craft tools.

The default value is just consistent with the rest of the APIs.  Worth reconsidering that too at some point.

PR #88, opened 2022-02-15
ci: add python 3.10 tests (CRAFT-745)

cmatsuoka:
Add Python 3.10 to GHA tests workflow ~~and tox~~.

-----

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -92,7 +92,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, macos-10.15, windows-2019]
-        python-version: [3.8, 3.9]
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -92,7 +92,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, macos-10.15, windows-2019]
-        python-version: [3.8, 3.9]
+        python-version: [3.8, 3.9, 3.10]
```
jugmac00, line 95: Needs to be "3.10" - as YAML converts it to a float otherwise.

cmatsuoka: Thanks. I added it to craft-parts but I forgot to do the same thing here.

sergiusens (approved):

PR #89, opened 2022-02-15
setup: update classifiers

cmatsuoka:
This is a follow-up to PR #88 (CRAFT-745).

-----


facundobatista (approved):
+1

setup.py
```diff
@@ -79,12 +79,13 @@
     author_email="snapcraft@lists.snapcraft.io",
     url="https://github.com/canonical/craft-providers",
     classifiers=[
-        "Development Status :: 2 - Pre-Alpha",
+        "Development Status :: 4 - Beta",
```
sergiusens, line 82: is beta to be removed once adopted by snapcraft?

cmatsuoka: My plan was to promote it to stable once snapcraft reaches production. We may need to fix issues revealed during the snapcraft tests.

PR #94, opened 2022-02-25
Fixed new pyright issues.

facundobatista:
Also improved Makefile to be explicit about what to test in each case.


mr-cal (approved):
Also looks good!

PR #98, opened 2022-03-08
lxd: pass id for host mount uid mapping

### LINKED ISSUE #97: Filesystem uid mapping incorrect in lxd instance
When mounting a host filesystem on a lxd instance craft-providers is mapping the id of the current user instead of the id of the owner of the directory being mounted. If an application (such as Snapcraft) tries to create a file inside its project directory mounted from the host that has a different owner, it will fail. This happens during Snapcraft spread tests in gce.

cmatsuoka:
Instead of always using the running user's uid for id mapping, allow
the value to be passed by the application so that the uid of the host
directory owner can be used in situations where the two values are
different. Default to the current user's uid to keep compatibility
with existing behavior.

CRAFT-893, fixes #97

-----

### DIFF
--- a/tests/integration/lxd/test_launcher.py
+++ b/tests/integration/lxd/test_launcher.py
@@ -16,6 +16,7 @@
 #
 
 import io
+import os
 import pathlib
 import subprocess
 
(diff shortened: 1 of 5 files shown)


tests/integration/lxd/test_launcher.py
```diff
@@ -212,6 +212,29 @@ def test_launch_map_user_uid_true(instance_name, tmp_path):
             instance.delete()
 
 
+def test_launch_map_user_uid_true_no_uid(instance_name, tmp_path):
```
mr-cal, line 215: These two tests are identical:
`test_launch_map_user_uid_true`
`test_launch_map_user_uid_true_no_uid`

Should the first test use the `uid` parameter?

mr-cal (approved):

tests/integration/lxd/test_launcher.py
cmatsuoka: Oops, good catch, the tests shouldn't be identical.

PR #105, opened 2022-03-29
lxd: in is_user_permitted() check if connecting to LXD works

simondeziel:
LXD doesn't always run as the `lxd` group as that is changeable
by the user (`snap set lxd daemon.group=adm`). As such, it is
better to check the socket to be writeable instead of hardcoding
a group.


sergiusens (approved):
thanks!!!

PR #107, opened 2022-03-30
refactor: instance_config.py

mr-cal:

-----
This is a pre-requisite for CRAFT-970 (caching snap revision installed in an environment).

`instance_config.py` tightly encapsulated the `compatibility_tag` parameter and was not readily usable for storing other data.

Improvements from refactoring:
1. Snap revisions can now be stored in the config file.
2. An existing config can now be updated, instead of just overwriting.
3. `test` and `cat` subprocess commands replaced with `executor.pull_file()`

(CRAFT-969)

### DIFF
--- a/craft_providers/bases/instance_config.py
+++ b/craft_providers/bases/instance_config.py
@@ -18,94 +18,128 @@
 """Persistent instance config / datastore resident in provided environment."""
 
 import io
-import logging
 import pathlib
-import subprocess
-from typing import Dict, Optional
+from typing import Any, Dict, Optional
 
 import pydantic
 import yaml
 
 from craft_providers import Executor, errors
+from craft_providers.util import temp_paths
 
 from .errors import BaseConfigurationError
 
-logger = logging.getLogger(__name__)
 
-
-class InstanceConfiguration(pydantic.BaseModel):
+class InstanceConfiguration(pydantic.BaseModel, extra=pydantic.Extra.forbid):
     """Instance configuration datastore.
 
     :param compatibility_tag: Compatibility tag for instance.
+
+    :param snaps: dictionary of snaps and their revisions, e.g.
+      snaps:
+        snapcraft:
+          revision: "x100"
+        charmcraft:
+          revision: 834
     """
 
-    compatibility_tag: str
+    compatibility_tag: Optional[str] = None
+    snaps: Optional[Dict[str, Dict[str, Any]]] = None
 
     @classmethod
-    def unmarshal(cls, data: Dict[str, str]) -> "InstanceConfiguration":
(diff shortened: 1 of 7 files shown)


craft_providers/bases/instance_config.py
```diff
@@ -18,94 +18,135 @@
...
+        """
+        with temp_paths.home_temporary_directory() as temp_directory:
+            temp_config_file = temp_directory / "craft-instance.conf"
+            try:
+                executor.pull_file(source=config_path, destination=temp_config_file)
+            except errors.ProviderError as error:
+                raise BaseConfigurationError(
+                    brief=f"Failed to read instance config in environment at {config_path}",
+                ) from error
+            except FileNotFoundError:
```
cmatsuoka, line 100: Doing the check/cat in a single operation was a good idea!

craft_providers/bases/instance_config.py
```diff
@@ -18,94 +18,135 @@
...
+                return None
+            with open(temp_config_file, encoding="utf8") as file:
+                data = yaml.safe_load(file)
+                if data is None:
+                    return None
+
+                return cls.unmarshal(data)
```
cmatsuoka, line 107: We can unmarshal to a variable and raise `BaseConfigurationError` here instead of raising it directly from `unmarshal`.

cmatsuoka (changes requested):
Looks good, added some notes about consistent definition of `unmarshal` (it's a pattern used in different projects) and other minor issues. I'd also copy the description of changes from the PR to the commit text for easier reference when browsing the git log.

craft_providers/bases/instance_config.py
mr-cal: Is this a general comment on the previous implementation?  Or a suggestion for this PR?

mr-cal: Or, a comment on the two exceptions from one try statement?

tests/unit/bases/test_buildd.py
```diff
@@ -26,6 +26,17 @@
 DEFAULT_FAKE_CMD = ["fake-executor"]
 
 
+@pytest.fixture()
+def mock_load():
+    with mock.patch(
```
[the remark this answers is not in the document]
mr-cal: Simplified.  `mock.patch` vs `mocker.patch` is a subtle difference.

craft_providers/bases/instance_config.py
cmatsuoka: Your version is better because it only makes a single call to execute things in the container.

cmatsuoka (approved):
Thanks Callahan!

PR #109, opened 2022-04-04
buildd: disable automatic APT in instance setup

facundobatista:
Disable automatic APT work during the BuildBase instance setup, so any
explicitly specified APT operation (in the setup itself or later) will
not fail because any automatic work has the cache lock.

--
CRAFT-958

### DIFF
--- a/craft_providers/bases/buildd.py
+++ b/craft_providers/bases/buildd.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2021 Canonical Ltd.
+# Copyright 2021-2022 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


craft_providers/bases/buildd.py
```diff
@@ -271,6 +266,32 @@ def setup(
...
+        includes apt work which may clash the triggered unattended jobs).
+
+        :param executor: Executor for target container.
+        :param deadline: Optional time.time() deadline.
+        """
+        _check_deadline(deadline)
+        # set the verification frequency in 10000 days and disable the upgrade
+        content = dedent(
+            """\
+            APT::Periodic::Update-Package-Lists "10000";
```
sergiusens, line 285: Doing this now would mean that me need to add craft-parts support to update the cache before fetching new things.

sergiusens (commented):
Looks good, I do have a concern around disabling the apt cache update in relation to package set changes in a project.

I also would appreciate the commit being amended to follow https://github.com/canonical/craft-providers#committing-code

facundobatista:
> Looks good, I do have a concern around disabling the apt cache update in relation to package set changes in a project.
> 
> I also would appreciate the commit being amended to follow https://github.com/canonical/craft-providers#committing-code

Done, thanks!

sergiusens (approved):

PR #110, opened 2022-04-05
actions: do not re-install same revision of snap on target

mr-cal:

-----
### Overview
Cache snap revision in the target's `InstanceConfiguration`.
If the revision on the target matches the revision on the host, then the snap will not be re-injected or re-installed on the target.

### Testing

- Snap is injected: **5.6 seconds**
- Store installed from store: **12.8 seconds**
- Snap injection/installation is skipped: **0.5 seconds**

These are informal results performed on a somewhat fast laptop.
I would expect an average charm developer to see build times cut by 5-15 seconds.

### Note on assertations
I ended up removing the `snap known` and `snap ack` calls from this PR.  I was struggling to get the code to work consistently and the unit testing was proving to be even more challenging.  

In retrospect, a better approach may have been to completely port over the data structures in `snapcraft/internal/repo/snaps.py` and `snapcraft/internal/build_providers/_snap.py` rather than add functionality piece-by-piece to the existing data structures.

(CRAFT-970)

### DIFF
--- a/craft_providers/actions/snap_installer.py
+++ b/craft_providers/actions/snap_installer.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2021 Canonical Ltd.
+# Copyright 2021-2022 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -23,12 +23,13 @@ import pathlib
 import shlex
 import subprocess
 import urllib.parse
-from typing import Iterator
+from typing import Iterator, Optional
 
 import requests
 import requests_unixsocket  # type: ignore
 
 from craft_providers import Executor
+from craft_providers.bases.instance_config import InstanceConfiguration
 from craft_providers.errors import ProviderError, details_from_called_process_error
 from craft_providers.util import snap_cmd, temp_paths
 
@@ -70,7 +71,7 @@ def _pack_host_snap(*, snap_name: str, output: pathlib.Path) -> None:
         "snap",
         "pack",
         f"/snap/{snap_name}/current/",
-        f"--filename={output.as_posix()}",
+        f"--filename={output}",
     ]
 
     logger.debug("Executing command on host: %s", shlex.join(cmd))
@@ -81,9 +82,47 @@ def _pack_host_snap(*, snap_name: str, output: pathlib.Path) -> None:
     )
 
 
+def _get_host_snap_revision(snap_name: str) -> str:
+    """Get the revision of the snap on the host."""
+    quoted_name = urllib.parse.quote(snap_name, safe="")
+    url = f"http+unix://%2Frun%2Fsnapd.socket/v2/snaps/{quoted_name}"
+    try:
+        snap_info = requests_unixsocket.get(url)
+    except requests.exceptions.ConnectionError as error:
+        raise SnapInstallationError(
+            brief="Unable to connect to snapd service."
+        ) from error
+    snap_info.raise_for_status()
+    return snap_info.json()["result"]["revision"]
+
+
+def _get_target_snap_revision(snap_name: str, executor: Executor) -> Optional[str]:
+    """Get the revision of the snap on the target."""
+    config = InstanceConfiguration.load(executor=executor)
+    if config is not None and config.snaps is not None:
+        snap = config.snaps.get(snap_name)
+        if snap is not None:
+            return str(snap.get("revision"))
+    return None
+
+
+def _get_store_snap_revision(snap_name: str, channel: str) -> str:
+    """Get the revision of a snap from the store."""
+    quoted_name = urllib.parse.quote(snap_name, safe="")
+    url = f"http+unix://%2Frun%2Fsnapd.socket/v2/find?name={quoted_name}"
+    try:
+        snap_info = requests_unixsocket.get(url)
+    except requests.exceptions.ConnectionError as error:
+        raise SnapInstallationError(
+            brief="Unable to connect to snapd service."
+        ) from error
+    snap_info.raise_for_status()
+    return snap_info.json()["result"][0]["channels"][channel]["revision"]
+
+
 @contextlib.contextmanager
 def _get_host_snap(snap_name: str) -> Iterator[pathlib.Path]:
-    """Get snap installed on host.
+    """Get snap installed on host containing the config.
 
     Snapd provides an API to fetch a snap. First use that to fetch a snap.
     If the snap is installed using `snap try`, it may fail to download. In
@@ -93,12 +132,11 @@ def _get_host_snap(snap_name: str) -> Iterator[pathlib.Path]:
     """
     with temp_paths.home_temporary_directory() as tmp_dir:
         snap_path = tmp_dir / f"{snap_name}.snap"
-
         try:
             _download_host_snap(snap_name=snap_name, output=snap_path)
         except SnapInstallationError:
             logger.warning(
-                "Failed to fetch snap from snapd, falling back to `snap pack` to recreate."
+                "Failed to fetch snap from snapd, falling back to `snap pack` to recreate"
             )
             _pack_host_snap(snap_name=snap_name, output=snap_path)
 
@@ -111,6 +149,21 @@ def inject_from_host(*, executor: Executor, snap_name: str, classic: bool) -> No
     :raises SnapInstallationError: on unexpected error.
     """
     target_snap_path = pathlib.Path(f"/tmp/{snap_name}.snap")
+    host_revision = _get_host_snap_revision(snap_name=snap_name)
+    target_revision = _get_target_snap_revision(snap_name=snap_name, executor=executor)
+    logger.debug(
+        "Revisions found for snap %r: host = %r, target = %r",
+        snap_name,
+        host_revision,
+        target_revision,
+    )
+
+    if target_revision is not None and target_revision == host_revision:
+        logger.debug(
+            "Skipping snap injection for snap %r: target is already up-to-date with revision on host",
+            snap_name,
+        )
+        return
 
     try:
         # Clean outdated snap, if exists.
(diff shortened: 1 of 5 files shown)


mr-cal:
Unit tests are passing locally on python 3.8 and locally in `tox`.  Not sure why it's failing only in the CI...

jugmac00:
> Unit tests are passing locally on python 3.8 and locally in `tox`. Not sure why it's failing only in the CI...

FWIW the unit tests also fail for me on Python 3.8/Ubuntu 20.04.

Depending on the state of your `tox` env it may help to rebuild it via `tox -e py38 -r`.

craft_providers/actions/snap_installer.py
```diff
@@ -111,6 +153,15 @@ def inject_from_host(*, executor: Executor, snap_name: str, classic: bool) -> No
     :raises SnapInstallationError: on unexpected error.
     """
     target_snap_path = pathlib.Path(f"/tmp/{snap_name}.snap")
+    host_revision = _get_host_snap_revision(snap_name=snap_name)
+    target_revision = _get_target_snap_revision(snap_name=snap_name, executor=executor)
+
+    if target_revision is not None and target_revision == host_revision:
+        logger.debug(
```
facundobatista, line 160: It may be useful to log which version that is.

Furthermore, in these cases I've found that when something weird happens around this mechanism, the first thing you need to understand the situation are the versions found in both places (host and target). So I would directly always log both values (in debug, of course, with repr) and then another line if skipped.

craft_providers/actions/snap_installer.py
```diff
@@ -139,6 +190,12 @@ def inject_from_host(*, executor: Executor, snap_name: str, classic: bool) -> No
             check=True,
             capture_output=True,
         )
+
+        InstanceConfiguration.update(
```
facundobatista, line 194: Is this `InstanceConfiguration.update` call expected to raise `subprocess.CalledProcessError`? If not, it should be outside the "try block"...

craft_providers/actions/snap_installer.py
```diff
@@ -189,6 +255,12 @@ def install_from_store(
...
+            executor=executor,
+            data={"snaps": {snap_name: {"revision": store_revision}}},
```
facundobatista, line 261: This `update` thing does not work like in dictionaries (took me several minutes of reading in pydantic's `main.py` code). It updates the first key, setting the rest.

I went forward and tested it:

```
(Pdb) InstanceConfiguration.update(executor=mock_executor, data={"snaps": {"some-snap-name": {"revision": 1234}}})
InstanceConfiguration(compatibility_tag=None, snaps={'some-snap-name': {'revision': 1234}})
(Pdb) InstanceConfiguration.update(executor=mock_executor, data={"snaps": {"other-snap-name": {"revision": 5678}}})
InstanceConfiguration(compatibility_tag=None, snaps={'other-snap-name': {'revision': 5678}})
```

This means that there's no point in having a "complex" structure below. And having a key called `snaps` is misleading because you cannot save more than one snap in this way.

Or am I missing something?

craft_providers/actions/snap_installer.py
```diff
@@ -81,9 +82,51 @@ def _pack_host_snap(*, snap_name: str, output: pathlib.Path) -> None:
...
+        ) from error
+    snap_info.raise_for_status()
+    return snap_info.json()["result"]["revision"]
+
+
+def _get_target_snap_revision(snap_name: str, executor: Executor) -> Optional[str]:
+    """Get the revision of the snap on the target."""
+    config = InstanceConfiguration.load(executor=executor)
+    if config is not None:
+        data = config.marshal()
```
facundobatista, line 103: Do you need this `marshal` call? you can access the info doing `data.snaps[snap_name]["revision"]`

craft_providers/actions/snap_installer.py
```diff
@@ -81,9 +82,51 @@ def _pack_host_snap(*, snap_name: str, output: pathlib.Path) -> None:
...
+    quoted_name = urllib.parse.quote(snap_name, safe="")
+    url = f"http+unix://%2Frun%2Fsnapd.socket/v2/find?name={quoted_name}"
+    try:
+        snap_info = requests_unixsocket.get(url)
+    except requests.exceptions.ConnectionError as error:
+        raise SnapInstallationError(
+            brief="Unable to connect to snapd service."
+        ) from error
+    snap_info.raise_for_status()
+    return snap_info.json()["result"][0]["channels"][f"{channel}"]["revision"]
```
facundobatista, line 124: why `f"{channel}"` if `channel` is already a `str`?

craft_providers/actions/snap_installer.py
```diff
@@ -165,6 +222,15 @@ def install_from_store(
     :raises SnapInstallationError: on unexpected error.
     """
     target_snap_path = pathlib.Path(f"/tmp/{snap_name}.snap")
+    store_revision = _get_store_snap_revision(snap_name=snap_name, channel=channel)
+    target_revision = _get_target_snap_revision(snap_name=snap_name, executor=executor)
+
+    if target_revision is not None and target_revision == store_revision:
```
facundobatista, line 228: Same detail about logging both revisions than above...

facundobatista (changes requested):
Hello! Great work! I annotated a couple of details inline... thanks!

craft_providers/actions/snap_installer.py
mr-cal: Good catch. I've updated the function to recurse through nested data.  I've also added 2 more unit tests for this.

craft_providers/actions/snap_installer.py
```diff
@@ -111,6 +149,21 @@ def inject_from_host(*, executor: Executor, snap_name: str, classic: bool) -> No
...
+    logger.debug(
+        "For snap %r: host revision = %r, target revision = %r",
+        snap_name,
+        host_revision,
+        target_revision,
+    )
+
+    if target_revision is not None and target_revision == host_revision:
+        logger.debug(
+            "Skipping snap injection for snap %r: target is already up-to-date with revision on host.",
```
sergiusens, line 163: I believe we aren't ending sentences for logs anymore
```suggestion
            "Skipping snap injection for snap %r: target is already up-to-date with revision on host",
```

mr-cal:
My laptop is running 21.10.  I setup a `lxc` container from the Ubuntu 20.04 image with Python 3.8.10.  The tox tests are still passing using `tox -e py38 -r`.

@jugmac00 - What version of python are you using, 3.8.10?  

Github CI runs 3.8.12.  I don't think the difference would cause it to fail, but I've been struggling to run that specific microversion with tox.

jugmac00:
> @jugmac00 - What version of python are you using, 3.8.10?

Yes :-/

> Github CI runs 3.8.12. I don't think the difference would cause it to fail, but I've been struggling to run that specific microversion with tox.

I think we never had a feature request to select a specific micro version, but what I usually do when I need to reproduce issues:

Run a docker container with the specific Python version (or install it), and then install `tox` with that Python version.

Then run `tox -e py` for the given project. This command uses the Python interpreter `tox` was installed with.

I uploaded the output of `tox -e py38 -rvv`  for you - if you'd like to compare that with your local output:
https://pastebin.canonical.com/p/ZHtPqSrr9H/

If you need me to run some commands or to give more details about my setup (ie lxd version or similar), just ping me here or on Mattermost (Europe friendly time). Or you can step through the code on my laptop (vs code live share or just Google meet).

craft_providers/actions/snap_installer.py
```diff
@@ -111,6 +149,21 @@ def inject_from_host(*, executor: Executor, snap_name: str, classic: bool) -> No
     :raises SnapInstallationError: on unexpected error.
     """
     target_snap_path = pathlib.Path(f"/tmp/{snap_name}.snap")
+    host_revision = _get_host_snap_revision(snap_name=snap_name)
+    target_revision = _get_target_snap_revision(snap_name=snap_name, executor=executor)
+    logger.debug(
+        "For snap %r: host revision = %r, target revision = %r",
```
facundobatista, line 155: nitpick (IMO shorter and more clear): "Revisions found for snap %r: host=%r, target=%r"

craft_providers/actions/snap_installer.py
```diff
@@ -189,8 +263,14 @@ def install_from_store(
             check=True,
             capture_output=True,
         )
+
```
facundobatista, line 266: here and above, it would be great if you get rid of this change (it's a blank think, it just dirties diffs)

craft_providers/bases/instance_config.py
```diff
@@ -118,6 +118,26 @@ def save(
             file_mode="0644",
         )
 
+    @classmethod
+    def _update_nested_dictionaries(
+        cls, config_data: Dict[str, Any], new_data: Dict[str, Any]
+    ) -> Dict[str, Any]:
```
facundobatista, line 124: It's weird that it changes the received object and then returns it. I would expect that it returns the result if the original is not modified, or viceversa.

Also, consider isolating this code (which is great) in a regular function outside the class.

facundobatista (changes requested):
Hello! I added a couple of small details. Thanks!

craft_providers/bases/instance_config.py
mr-cal: I moved the code outside the class in the same file.

I agree that modifying a python dictionary and returning it is somewhat odd.  Based on your comment, I tried conditionally returning the result when the original dictionary is modified.  Between the extra code and the extra type checking to satisfy the linters, the function got more complex than it needed to be.  So I reverted.

I checked on stack overflow for a better solution, most solutions seem to be following a similar paradigm.

facundobatista: Do not change the behaviour conditionally. Or *always* change the received dict and return `None`, or *always* don't touch the received dict and return the new object. I'd go for the former (and note the behaviour in the docstring).

facundobatista (approved):
Great! Let's go on.

PR #111, opened 2022-04-05
buildd: just warm up already existing instances

facundobatista:
The full setup takes too long and is not needed, so I created a new warmup
method for when the instance already exists (and was fully setup in the
past).


mr-cal (approved):
Looks good to me!

sergiusens (approved):

PR #123, opened 2022-05-20
tests: remove snapcraft:core

sergiusens:
core has been removed by multipass, craft-providers does not need to
support core so remove the test for it.

-----


facundobatista (approved):
+1

PR #128, opened 2022-06-17
lint: move .pylintrc to pyproject.toml

mr-cal:

-----

Follow `pylint` config style used in `snapcraft` and `craft-parts`.  

I prefer this for a few reasons:
1. Consolidates multiple configs into one place
2. Marginally less prone to breakage
3. Easier to work with (19 vs 524 lines of code)


facundobatista (changes requested):
Mmm... why are you changing the config values? e.g. max locals from 16 to 8, or line length from 100 to 88...

mr-cal:
> Mmm... why are you changing the config values? e.g. max locals from 16 to 8, or line length from 100 to 88...

Changing `line length` to 88 is to match this project's config for flake8 and isort.  (I don't know who chose 88, but that appears to be the preference in `snapcraft` and `craft-parts` too)

Changing `max locals` from 16 to 18 was a mistake, I reverted that.

Changing `min similar lines` from 12 to 13 was also a mistake.  Reverted.

PR #134, opened 2022-08-03
lxd: use LXD-compatible instance names

mr-cal:

-----
### Overview
The provided name for a LXD executor object is converted to comply with [LXD naming conventions](https://linuxcontainers.org/lxd/docs/master/instances/) for instances.

### Problem
LXC has a 63 character limit on instance names.  [This snapcraft commit](https://github.com/snapcore/snapcraft/commit/21cb1ab9ae26cf3322356c190eb555ed05f2d498) adds 18-20 characters to the image name.  This is blocking snaps with long names from upgrading to core22.

Charmcraft has the [same issue](https://github.com/canonical/charmcraft/issues/629).

#### Snapcraft naming convention
```
snapcraft-<project-name>-on-<build-on>-for-<build-for>-<project-path-inode>
└  10   ┘└   2 - 41     ┘└   8-11    ┘└    9-12       ┘└       10         ┘
```
Min size: 10 + 2 + 8 + 9 + 10 = 39
Max size: 10 + 41 + 11 + 12 + 10 = 84

#### Charmcraft naming convention
```
charmcraft-<project-name>-<project-path-inode>-<bases-index>-<build-on-index>-<build-for>
└  11   ┘└    2 - ∞      ┘└       10          ┘└     2      ┘└      2        ┘└    7    ┘
```
Min size:   11 + 2 + 10 + 2 + 2 + 7 = 34
Max size: unbound

### Solution
`craft-providers` is responsible for complying with each provider’s specific requirements.
For LXD, `craft-providers` will:
1. truncate the first 40 characters name
2. remove forbidden characters
3. append a hash of the name
```
<truncated-name>-<hash-of-name>
└    1 - 40    ┘1└     20     ┘
```
Min size: 1 + 1 + 20 = 22
Max size: 40 + 1 + 20 = 61

Note: This is an internal change.  Applications using `craft-providers` should continue to choose their own unique names for instances.  The only restriction is that the name must contain at least one alphanumeric character.

LP: #[1981205](https://bugs.launchpad.net/snapcraft/+bug/1981205)
(CRAFT-1235)

### DIFF
--- a/craft_providers/lxd/lxd_instance.py
+++ b/craft_providers/lxd/lxd_instance.py
@@ -113,19 +113,23 @@ class LXDInstance(Executor):
                 details="name must contain at least one alphanumeric character",
             )
 
-        # truncate to 40 characters
-        truncated_name = trimmed_name.group("valid_part_of_name")[:40]
+        # if the original name meets LXD's naming convention, then use the original name
+        if (
+            self.name == trimmed_name.group("valid_part_of_name")
+            and len(self.name) <= 63
+        ):
+            instance_name = self.name
 
-        # if the name is unchanged, then use the original name
-        if self.name == truncated_name:
-            self.instance_name = self.name
+        # else, continue converting the name
         else:
+            # truncate to 40 characters
+            truncated_name = trimmed_name.group("valid_part_of_name")[:40]
             # hash the entire name, not the truncated name
-            hasher = hashlib.sha1()
-            hasher.update(self.name.encode())
-            hashed_name = hasher.hexdigest()[:20]
+            hashed_name = hashlib.sha1(self.name.encode()).hexdigest()[:20]
+            instance_name = f"{truncated_name}-{hashed_name}"
 
-            self.instance_name = f"{truncated_name}-{hashed_name}"
+        self.instance_name = instance_name
+        logger.debug("Set LXD instance name to %s")
 
     def _finalize_lxc_command(
         self,
--- a/craft_providers/lxd/lxd_instance.py
+++ b/craft_providers/lxd/lxd_instance.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2021 Canonical Ltd.
+# Copyright 2021-2022 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -17,10 +17,12 @@
 
 """LXD Instance Executor."""
 
+import hashlib
 import io
 import logging
 import os
 import pathlib
+import re
 import shutil
 import subprocess
 import tempfile
@@ -48,6 +50,17 @@ class LXDInstance(Executor):
         remote: str = "local",
         lxc: Optional[LXC] = None,
     ):
+        """Create an LXD executor.
+
+        To comply with LXD naming conventions, the supplied name is converted to a
+        LXD-compatible name before creating the instance.
+
+        :param name: Unique name of the lxd instance
+        :param default_command_environment: command environment
+        :param project: name of lxd project
+        :param remote: name of lxd remote
+        :param lxc: LXC instance object
+        """
         super().__init__()
 
         if default_command_environment is not None:
@@ -56,6 +69,7 @@ class LXDInstance(Executor):
             self.default_command_environment = {}
 
         self.name = name
+        self._set_instance_name()
         self.project = project
         self.remote = remote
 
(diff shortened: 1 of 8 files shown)


craft_providers/lxd/lxd_instance.py
```diff
@@ -64,6 +78,51 @@ def __init__(
...
+
+        # truncate to 40 characters
+        truncated_name = trimmed_name.group("valid_part_of_name")[:40]
+
+        # hash the entire name, not the truncated name
+        hasher = hashlib.sha1()
+        hasher.update(self.name.encode())
+        hashed_name = hasher.hexdigest()[:20]
+
+        self.instance_name = f"{truncated_name}-{hashed_name}"
```
cmatsuoka, line 124: Shouldn't we use the new instance name only if the old-style name is invalid or too large, to keep using existing instances?

mr-cal: Good catch, I added that in.

craft_providers/lxd/lxd_instance.py
```diff
@@ -64,6 +78,55 @@ def __init__(
...
+            raise LXDError(
+                brief=(f"failed to create LXD instance with name {self.name!r}."),
+                details="name must contain at least one alphanumeric character",
+            )
+
+        # truncate to 40 characters
+        truncated_name = trimmed_name.group("valid_part_of_name")[:40]
+
+        # if the name is unchanged, then use the original name
+        if self.name == truncated_name:
```
cmatsuoka, line 120: This test doesn't seem correct, truncated names are clipped at 40 char size but previous valid names could have up to 64?

craft_providers/lxd/lxd_instance.py
```diff
@@ -64,6 +78,55 @@ def __init__(
...
+            self.instance_name = self.name
```
facundobatista, line 121: I think we should do this validation before truncating.

If the proposed name by the app is valid and, say, 55c long, it should be ok and use it.

So, removing non valid chars and trimming digits and dashes it's a mandatory step (so we avoid invalid names), but then if the len of the resulting thing is < 62, just use it! if it's > 62, truncate to 40ish and append hash, yes.

What do you think?

craft_providers/lxd/lxd_instance.py
```diff
@@ -64,6 +78,55 @@ def __init__(
...
+        else:
+            # hash the entire name, not the truncated name
+            hasher = hashlib.sha1()
+            hasher.update(self.name.encode())
+            hashed_name = hasher.hexdigest()[:20]
```
facundobatista, line 126: No need for the "update" part (which is more useful when you don't have the full content when you start). IOW, you can just pass the encoded name when creating the hasher...

```
>>> hashlib.sha1(b"asd").hexdigest()
'f10e2821bbbea527ea02200352313bc059445190'
```

tests/unit/lxd/test_lxd_instance.py
```diff
@@ -841,10 +864,107 @@ def test_unmount_error(mock_lxc, instance):
...
+        ("test--", "test"),
+        ("test1-", "test1"),
+        # trim away invalid characters
+        ("test$", "test"),
+        ("test-!@#$%^&*()test", "test-test"),
+        ("$1test", "test"),
+        ("test-$", "test"),
+        # truncate at 40 characters
+        (
+            "this-is-41-characters-xxxxxxxxxxxxxxxxxXx",
```
facundobatista, line 915: this is what I meant above, this is a totally valid name!

craft_providers/lxd/lxd_instance.py
```diff
@@ -64,6 +78,55 @@ def __init__(
         else:
             self.lxc = lxc
 
+    def _set_instance_name(self) -> None:
```
facundobatista, line 81: Probably this function should log (in debug) the final name, converted or not! because developers may want to "discover" the name of the instance for cases when there is need to handle them directly through LXC. What do you think?

facundobatista (changes requested):
Great work!! Added some details and a couple of questions...

craft_providers/lxd/lxd_instance.py
mr-cal: Yep, great idea! Added logging

tests/unit/lxd/test_lxd_instance.py
mr-cal: Whoops, thanks for catching.  This was an accidental oversight.  Fixed!

craft_providers/lxd/lxd_instance.py
mr-cal: Whoops, thanks for catching this too.  Fixed!

craft_providers/lxd/lxd_instance.py
```diff
@@ -64,6 +78,59 @@ def __init__(
...
+
+        # trim digits and dashes from the beginning and dashes from the end
+        trimmed_name = re.compile(r"^[0-9-]*(?P<valid_part_of_name>.*?)[-]*$").search(
+            name_with_valid_chars
+        )
+        if not trimmed_name or trimmed_name.group("valid_part_of_name") == "":
+            raise LXDError(
+                brief=(f"failed to create LXD instance with name {self.name!r}."),
+                details="name must contain at least one alphanumeric character",
+            )
```
facundobatista, line 114: Let's isolate the rest of the code from this regex, just do

```
valid_name = trimmed_name.group("valid_part_of_name")
```

... and use that name after this line.

craft_providers/lxd/lxd_instance.py
```diff
@@ -64,6 +78,59 @@ def __init__(
...
+        # else, continue converting the name
+        else:
+            # truncate to 40 characters
+            truncated_name = trimmed_name.group("valid_part_of_name")[:40]
+            # hash the entire name, not the truncated name
+            hashed_name = hashlib.sha1(self.name.encode()).hexdigest()[:20]
+            instance_name = f"{truncated_name}-{hashed_name}"
+
+        self.instance_name = instance_name
+        logger.debug("Set LXD instance name to %s")
```
facundobatista, line 132: to what? you're missing something to replace the `%s`... also you probably should use `%r` here.

btw, you're not testing this log call, right?

facundobatista (changes requested):
two small things!

craft_providers/lxd/lxd_instance.py
mr-cal: Added the variable and a test.  Today I learned `assert Exact(...) in logs.debug`

craft_providers/lxd/lxd_instance.py
mr-cal: +1

mr-cal:
Changing the instance name has uncovered the next issue: `hostname`.  The `hostname` has a similar 63 max character limitation, but it's configured by the Base object, not the Executor object.

Previously, packing a charm with the name `this-is-35-characters-xxxxxxxxxxxxx` produced this error:
```
2022-08-05 15:47:58.902 * Command exit code: 1
2022-08-05 15:47:58.902 * Command output: b'Creating charmcraft-this-is-35-characters-xxxxxxxxxxxxx-21105008-0-0-amd64\n'
2022-08-05 15:47:58.902 * Command standard error output: b'Error: Invalid instance name: Name must be 1-63 characters long\n
```

Using this PR, charmcraft produces this error:
```
2022-08-05 15:48:36.933 Failed to set hostname.
* Command that failed: 'lxc --project charmcraft exec local:charmcraft-this-is-35-characters-xxxxxxx-95e3d232132477219c4f -- env PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin CHARMCRAFT_MANAGED_MODE=1 hostname -F /etc/hostname'
* Command exit code: 1
* Command standard error output: b'hostname: name too long\n'
```

Snapcraft has the same behavior.  I haven't thought of a solution that doesn't involve a breaking change, but I'll test it more.

facundobatista (approved):
Wonderful, thanks!

mr-cal:
Thanks for the approval!
I'm going to hold this PR as a draft, because it is breaking the command `*craft clean` in all craft applications.
Once that issue is solved, I will land this PR.

PR #136, opened 2022-08-08
bases: ensure BuilddBase hostname is valid

mr-cal:

-----

This PR ensures that the `hostname` is valid, which was previously raising an error.

`BuilddBase`'s `hostname` parameter has similar naming conventions to [LXD instance names](https://github.com/canonical/craft-providers/pull/134).  Yet, the naming convention is different enough that I didn't combine the functionality between the two.

(CRAFT-1234)

### DIFF
--- a/tests/unit/bases/test_buildd.py
+++ b/tests/unit/bases/test_buildd.py
@@ -19,6 +19,7 @@ from textwrap import dedent
 from unittest import mock
 
 import pytest
+from logassert import Exact  # type: ignore
 from pydantic import ValidationError
 
 from craft_providers.bases import (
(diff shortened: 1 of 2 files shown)


facundobatista (approved):
Gooooooood!

tests/unit/bases/test_buildd.py
```diff
@@ -855,3 +856,84 @@ def test_warmup_never_network(fake_process, fake_executor, mock_load):
...
+        ("test-$", "test"),
+        # this name contains invalid characters so it gets converted, even
+        # though it is 63 characters
+        (
+            "this-is-63-characters-with-invalid-characters-$$$xxxxxxxxxxxxxX",
+            "this-is-63-characters-with-invalid-characters-xxxxxxxxxxxxxX",
+        ),
+        # this name is longer than 63 characters, so it gets converted
+        (
+            "this-is-70-characters-with-valid-characters-xxxxxxxxxxxxxxxxxxXxxxxxxx",
```
cmatsuoka, line 906: This test could have the smallest size that gets trimmed (64) to ensure we're trimming at the right length.

cmatsuoka (approved):
Looks good, just added a small test comment.

tests/unit/bases/test_buildd.py
mr-cal: Good idea, added!

PR #140, opened 2022-08-16
executor: add exists and delete function to abstract base class

mr-cal:

-----

Add 2 functions, `exists()` and `delete()` to the `Executor` abstract base class.

This is required so applications can use these functions in the child classes.

(CRAFT-1251)

### DIFF
--- a/craft_providers/executor.py
+++ b/craft_providers/executor.py
@@ -120,3 +120,14 @@ class Executor(ABC):
         :param group: File owner group.
         :param user: File owner user.
         """
+
+    @abstractmethod
+    def delete(self) -> None:
+        """Delete instance."""
+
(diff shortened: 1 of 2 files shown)


craft_providers/executor.py
```diff
@@ -120,3 +120,14 @@ def push_file_io(
         :param group: File owner group.
         :param user: File owner user.
         """
+
+    @abstractmethod
+    def delete(self) -> None:
+        """Delete instance."""
+
+    @abstractmethod
+    def exists(self) -> bool:
```
facundobatista, line 129: Exists self? So philosophical!!

facundobatista (approved):
+1

PR #143, opened 2022-08-19
base: add optional list of system packages to install

mr-cal:

-----

### Overview
New parameter in `BuilddBase` class that installs system packages in the base.

### Details
I manually tested with `snapcraft` and it appears to be 100% compatible.  Currently, `snapcraft` is the only application installing additional system packages.

Similar to https://github.com/canonical/craft-providers/pull/137.

(CRAFT-1223)


facundobatista (approved):
Nice! +1

PR #145, opened 2022-08-24
bases: allow overriding of compatibility tag

mr-cal:

-----
New optional parameter `compatibility_tag` for `BuilddBase`.

This allows applications to easily override the default compatibility tag without changing the existing behavior.

Manually tested and confirmed on all *craft applications!

(CRAFT-1248)


facundobatista (approved):
Good!! +1

PR #147, opened 2022-08-30
bases: fix bug in BuilddBase hostname validation logic

mr-cal:

-----
I made a mistake in hostname validation logic!  

This fixes a bug where hostnames longer than 64 characters may not have trailing hyphens removed.

### DIFF
--- a/craft_providers/bases/buildd.py
+++ b/craft_providers/bases/buildd.py
@@ -185,8 +185,11 @@ class BuilddBase(Base):
         :raises BaseConfigurationError: if the hostname contains no
           alphanumeric characters
         """
+        # truncate to 63 characters
+        truncated_name = hostname[:63]
+
         # remove anything that is not an alphanumeric character or hyphen
-        name_with_valid_chars = re.sub(r"[^\w-]", "", hostname)
+        name_with_valid_chars = re.sub(r"[^\w-]", "", truncated_name)
 
         # trim hyphens from the beginning and end
         trimmed_name = re.compile(r"^[-]*(?P<valid_name>.*?)[-]*$").search(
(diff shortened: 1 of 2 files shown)


tests/unit/bases/test_buildd.py
```diff
@@ -1129,6 +1129,16 @@ def test_set_hostname_unchanged(hostname, logs):
...
             "this-is-64-characters-with-valid-characters-xxxxxxxxxxxxxxxxxxX",
         ),
+        # trim away away invalid characters and truncate to 63 characters
+        (
+            "-this-is-64-characters-and-has-invalid-characters-$$$xxxxxxxxxx-",
+            "this-is-64-characters-and-has-invalid-characters-xxxxxxxxxx",
+        ),
+        # ensure invalid ending characters are removed after truncating
+        (
+            "this-is-64-characters-and-has-a-hyphen-at-character-63-xxxxxxx-x",
```
mr-cal, line 1139: This was the failing scenario

craft_providers/bases/buildd.py
```diff
@@ -199,11 +202,8 @@ def _set_hostname(self, hostname: str) -> None:
             )
         valid_name = trimmed_name.group("valid_name")
```
facundobatista, line 203: Can we replace this with
```
valid_name = name_with_valid_chars.strip("-")
if not valid_name:
    raise BaseConfigurationError(...)
```
?

facundobatista (approved):
+1, but in the spirit of simplifying so the whole sequence is more understable, what about the change I proposed inline?

Thanks!

craft_providers/bases/buildd.py
mr-cal: 🤦 of course there was a more pythonic way to do this!  Simplified.

PR #149, opened 2022-09-08
buildd: set snapd http-proxy and https-proxy

mr-cal:

-----

If the environmental variables `http-proxy` or `https-proxy` are set, then they are configured with `snapd`.

This code was ported from [`snapcraft_legacy`](https://github.com/snapcore/snapcraft/blob/4a66b3b1b7e0cd39ba06dd4ab1c6d21c8cddd969/snapcraft_legacy/internal/build_providers/_base_provider.py#L510).

(CRAFT-1335)

### DIFF
--- a/craft_providers/bases/buildd.py
+++ b/craft_providers/bases/buildd.py
@@ -350,6 +350,7 @@ class BuilddBase(Base):
         )
         self._setup_apt(executor=executor, deadline=deadline)
         self._setup_snapd(executor=executor, deadline=deadline)
+        self._setup_snapd_proxy(executor=executor, deadline=deadline)
         self._install_snaps(executor=executor, deadline=deadline)
 
     def warmup(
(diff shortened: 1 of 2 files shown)


craft_providers/bases/buildd.py
```diff
@@ -755,6 +755,23 @@ def _setup_snapd(
                 capture_output=True,
                 check=True,
             )
+
+            _check_deadline(deadline)
+            http_proxy = self.environment.get("http_proxy")
```
sergiusens, line 760: if we only setup snapd once; a rerun should update proxy settings, so this might need to live somewhere else

sergiusens (approved):

craft_providers/bases/buildd.py
mr-cal: I split this code out and added it to `warmup()`

PR #151, opened 2022-09-15
snap installer: do not remove previous snap file

mr-cal:

-----
#### Overview
Do not remove the previous snap file (i.e. `/tmp/<snap-name>.snap`) before pushing a new snap file into the instance.

Pushing a file into an instance overwrites the previous file, so there is no need to delete the previous file.

#### Reasons to remove
1. This isn't done in `snapcraft_legacy/internal/build_providers`
2. It doesn't need to be done
3. It adds time and complexity

(CRAFT-1370)


facundobatista (approved):
+1

PR #152, opened 2022-09-15
snap installer: refactor

mr-cal:
- [X] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

-----
#### Overview

Minor refactor of `snap_installer`.
https://github.com/canonical/craft-providers/pull/151 needs to be reviewed and merged before this PR.

#### Details
No functional changes.  This is laying groundwork for the next PR to assert snaps.

Notable changes:
1. New integration test to inject a dangerously installed snap
2. Change `get_host_snap_revision()` to a generic function `get_host_snap_info()`, since I'll soon be needing more data than just the revision.
---
CRAFT-1371

PR #153, opened 2022-09-16
snap installer: assert injected snaps

mr-cal:

-----
### Overview
Assert signed snaps injected into a provider.

### Details
If a snap on the host is signed (not dangerous), then craft-providers will:
1. collect the assertions from the host with `snap known`
2. push the assertions to the instance
3. add the assertions in the instance with `snap ack`

If the snap is unsigned on the host, then the snap is installed dangerously in the instance.  This behavior is unchanged.

The implementation in [snapcraft_legacy](https://github.com/snapcore/snapcraft/blob/433a09262443d645acbae5204b5c9a44433f4b83/snapcraft_legacy/internal/build_providers/_snap.py) was used as a reference.

I tested this manually with snapcraft, switching between dangerous and non-dangerous snaps.  The integration tests here already test for both scenarios too (I added those tests in the [previous PR](https://github.com/canonical/craft-providers/pull/152)).

(CRAFT-1372)

### DIFF
--- a/craft_providers/actions/snap_installer.py
+++ b/craft_providers/actions/snap_installer.py
@@ -24,7 +24,7 @@ import pathlib
 import shlex
 import subprocess
 import urllib.parse
-from typing import Dict, Iterator, Optional
+from typing import Dict, Iterator, List, Optional
 
 import requests
 import requests_unixsocket  # type: ignore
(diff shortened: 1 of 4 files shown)


craft_providers/actions/snap_installer.py
```diff
@@ -180,6 +180,94 @@ def _get_host_snap(snap_name: str) -> Iterator[pathlib.Path]:
...
 
+def _get_assertion(query: List[str]) -> bytes:
+    """Get an assertion from snapd.
+
+    :param query: assertion query to pass to `snap known`
+    :returns: assertion data
+    :raises SnapInstallationError: if 'snap known' call fails
+    """
+    command = snap_cmd.formulate_known_command(query=query)
+    logger.debug("Executing command on host: %s", shlex.join(command))
```
facundobatista, line 191: Just log "command". Yes, it's ugly in the sense of "a list with strings", but when a behaviour is weird and you have to debug it, having *exactly* what you executed is better.

craft_providers/actions/snap_installer.py
```diff
@@ -180,6 +180,94 @@ def _get_host_snap(snap_name: str) -> Iterator[pathlib.Path]:
...
+    :param snap_id: ID of the snap
+    :param snap_revision: Revision of the snap
+
+    :yields: path to temporary assertion file
+    """
+    logger.debug("Creating an assert file for snap %r", snap_name)
+    assertion_queries = [
+        [
+            "account-key",
+            "public-key-sha3-384=BWDEoaqyr25nF5SNCvEv2v"
```
facundobatista, line 217: Where this hash comes from?

craft_providers/actions/snap_installer.py
```diff
@@ -180,6 +180,94 @@ def _get_host_snap(snap_name: str) -> Iterator[pathlib.Path]:
...
+            "7QnM9QsfCc0PBMYD_i2NGSQ32EF2d4D0hqUel3m8ul",
+        ],
+        ["snap-declaration", f"snap-name={snap_name}"],
+        ["snap-revision", f"snap-revision={snap_revision}", f"snap-id={snap_id}"],
+    ]
+
+    with temp_paths.home_temporary_file() as assert_file_path:
+        assert_file = open(assert_file_path, "wb")
```
facundobatista, line 225: Where this file is closed?

craft_providers/actions/snap_installer.py
```diff
@@ -180,6 +180,94 @@ def _get_host_snap(snap_name: str) -> Iterator[pathlib.Path]:
...
+def _get_assertions_file(
+    snap_name: str, snap_id: str, snap_revision: str
+) -> Iterator[pathlib.Path]:
+    """Get an assertion file for a snap.
+
+    :param snap_name: Name of snap to inject
+    :param snap_id: ID of the snap
+    :param snap_revision: Revision of the snap
+
+    :yields: path to temporary assertion file
```
facundobatista, line 211: this docstring is confusing here, as this is not a "generator that yields values", but a "context manager that will set the temporary assertion file as the target"... right?

tests/unit/actions/test_snap_installer.py
```diff
@@ -168,6 +179,65 @@ def test_inject_from_host_strict(
     fake_process,
     logs,
     tmp_path,
+):
+    # register 'snap known' calls
+    for _ in range(3):
+        fake_process.register_subprocess(["snap", "known", fake_process.any()])
```
sergiusens, line 185: why is this needed? I admit I am not as familiar with fake_process

sergiusens: Ah, I think I get it now

sergiusens (commented):
Looks good, I have one question. And did you try with using an existing instance created from a released snapcraft and then moving to one using this?

facundobatista (changes requested):
Good work!! I annotated a small detail and asked a couple of questions. Thanks!!

mr-cal:
> Looks good, I have one question. And did you try with using an existing instance created from a released snapcraft and then moving to one using this?

Yes, I tested this scenario too!

sergiusens (approved):

facundobatista (approved):
Awesome, thanks!

PR #154, opened 2022-09-21
executor: add mount method to base Executor class

mr-cal:

-----

Add `mount()` method to base Executor class.  This is required so applications can use these functions in the child classes.  This is similar to https://github.com/canonical/craft-providers/pull/140.

LXD's mount method was renamed to make room for new inherited `mount` method while still allowing access to the old method.

`LXDInstance.mount(host_source, target, device_name)` is now `LXDInstance.mount_with_device_name(host_source, target, device_name)`

None of the *craft applications use the `device_name` parameter, so there should be no impact.  

First usage will be in [rockcraft](https://github.com/canonical/rockcraft/pull/82).

(CRAFT-1383)

### DIFF
--- a/craft_providers/lxd/lxd_instance.py
+++ b/craft_providers/lxd/lxd_instance.py
@@ -414,12 +414,7 @@ class LXDInstance(Executor):
             remote=self.remote,
         )
 
-    def mount(
-        self,
-        *,
-        host_source: pathlib.Path,
-        target: pathlib.PurePath,
-    ) -> None:
+    def mount(self, *, host_source: pathlib.Path, target: pathlib.PurePath) -> None:
         """Mount host source directory to target mount point.
 
         Checks first to see if already mounted.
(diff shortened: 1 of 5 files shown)


craft_providers/lxd/lxd_instance.py
```diff
@@ -419,10 +419,32 @@ def mount(
...
 
+        Checks first to see if already mounted.
+        The source will be mounted as a disk named "disk-{target.as_posix()}".
+
+        :param host_source: Host path to mount.
+        :param target: Instance path to mount to.
+
+        :raises LXDError: On unexpected error.
+        """
+        self.mount_with_device_name(
```
sergiusens, line 433: is there a reason to not impement mount_with_device_name here?

mr-cal: Because the concept of a `device_name` doesn't exist for Multipass.

sergiusens: I think you did not understand me, I am not asking to change the signature of mount, but instead, in this LXD specific implementation instead of calling self.mount_with_device_name, just implement that here

tests/unit/conftest.py
```diff
@@ -119,6 +119,9 @@ def delete(self) -> None:
     def exists(self) -> bool:
         return True
 
+    def mount(self, *, host_source: pathlib.Path, target: pathlib.Path) -> None:
+        return
```
sergiusens, line 123: don't we have tests similar to what excercises push_file/pull_file which would leverage having a similar records structure?

craft_providers/lxd/lxd_instance.py
mr-cal: I think I understand and am now calling the underlying lxc calls instead.

mr-cal: Third try!  Now I have removed the `mount_with_device_name()` method.

sergiusens (approved):

PR #158, opened 2022-09-28
Avoid remote adding race condition with other processes (CRAFT-1043).

### LINKED ISSUE #96: Avoid race conditions when adding remotes in `configure_buildd_image_remote`
The function `configure_buildd_image_remote` in `craft_providers/lxd/remotes.py` should support being executed in parallel.

Currently there are race conditions where the check to see if `BUILDD_REMOTE_NAME in lxc.remote_list()` fails, but then other process creates it, and the `remote_add` call two lines below fails.

We could do something like the following:

```
    if BUILDD_REMOTE_NAME in lxc.remote_list():
        logger.debug("Remote %r already exists.", BUILDD_REMOTE_NAME)
    else:
        try:
            lxc.remote_add(
                remote=BUILDD_REMOTE_NAME,
                addr=BUILDD_REMOTE_ADDR,
                protocol="simplestreams",
            )
            logger.debug("Remote %r was successfully added.", BUILDD_REMOTE_NAME)
        except Exception:
            # re-checking in case the error comes from a creation race condition:
            if BUILDD_REMOTE_NAME not in lxc.remote_list():
                # nop, expose the problem
                raise

    return BUILDD_REMOTE_NAME
```

facundobatista:
Fixes #96.

### DIFF
--- a/craft_providers/lxd/remotes.py
+++ b/craft_providers/lxd/remotes.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2021 Canonical Ltd.
+# Copyright 2021-2022 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


mr-cal (approved):
+1

PR #159, opened 2022-09-28
snap installer: get assertion for the publisher's account id

mr-cal:

-----
### Overview
Adds another assertion when injecting a snap into a provider.

### Details
@tigarmo and I saw this problem with revision 309 of rockcraft.  The snap's publisher was unverified and produced the error:
```
error: cannot assert: assert failed: cannot resolve prerequisite assertion: account (pMLNwd28ezFdozetvrAOlLj3UqC9HKpe)
```

Looking into `rockcraft`'s snap info, it contains a publisher with a matching account id:
```
'publisher': {
  'id': 'pMLNwd28ezFdozetvrAOlLj3UqC9HKpe',
  'username': 'sergiusens',
  'display-name': 'Sergio Schvezov',
  'validation': 'unproven'
}
```
I assume the key `'validation': 'unproven'` is causing the error.

To resolve this, `craft-providers` now collects the publisher account with `snap known account account-id=pMLNwd28ezFdozetvrAOlLj3UqC9HKpe`

The assertion looks like this:
```
type: account
authority-id: canonical
revision: 223
account-id: pMLNwd28ezFdozetvrAOlLj3UqC9HKpe
display-name: Sergio Schvezov
timestamp: 2022-09-23T08:30:19.896972Z
username: sergiusens
validation: unproven
sign-key-sha3-384: BWDEoaqyr25nF5SNCvEv2v7QnM9QsfCc0PBMYD_i2NGSQ32EF2d4D0hqUel3m8ul

...
```

It appears to be a working and valid solution for the current rockcraft snap.  There are no regressions for snapcraft or charmcraft.

### DIFF
--- a/craft_providers/actions/snap_installer.py
+++ b/craft_providers/actions/snap_installer.py
@@ -24,7 +24,7 @@ import pathlib
 import shlex
 import subprocess
 import urllib.parse
-from typing import Dict, Iterator, List, Optional
+from typing import Any, Dict, Iterator, List, Optional
 
 import requests
 import requests_unixsocket  # type: ignore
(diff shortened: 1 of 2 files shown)


tigarmo (approved):
many thanks for looking into this!

PR #175, opened 2022-11-10
buildd: avoid getent hangs

facundobatista:
Also removed several `capture_output=True` for the cases where the captured output was actually not used.

### DIFF
--- a/craft_providers/bases/buildd.py
+++ b/craft_providers/bases/buildd.py
@@ -514,7 +514,6 @@ class BuilddBase(Base):
             _check_deadline(deadline)
             executor.execute_run(
                 ["apt-get", "update"],
-                capture_output=True,
                 check=True,
             )
         except subprocess.CalledProcessError as error:
@@ -532,7 +531,6 @@ class BuilddBase(Base):
             _check_deadline(deadline)
             executor.execute_run(
                 ["apt-get", "install", "-y"] + packages_to_install,
-                capture_output=True,
                 check=True,
             )
         except subprocess.CalledProcessError as error:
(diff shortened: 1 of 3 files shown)


craft_providers/bases/buildd.py
```diff
@@ -828,13 +816,19 @@ def _setup_wait_for_network(
...
-            )
-            if proc.returncode == 0:
-                return
+            try:
+                proc = executor.execute_run(
+                    ["getent", "hosts", "snapcraft.io"],
+                    check=False,
+                    timeout=1,
+                )
+            except subprocess.TimeoutExpired:
```
mr-cal, line 825: I have a question about this:
> we seen too many times that [getent] "hangs"

Are you talking about `getent` hanging as in "stuck forever"?

I have been observing the issue where `getent` doesn't get stuck (it times itself out after 5 seconds), but the `while True` loop in this method runs forever and `getent` gets called again and again.  I believe this would be solved with using `deadline`.

I was trying to simulate the failure with charmcraft.  With or without this PR, the behavior is the same - charmcraft stays in the while loop forever with no user feedback (unless running with a high verbosity level).

craft_providers/bases/errors.py
```diff
@@ -26,6 +26,15 @@ class BaseConfigurationError(ProviderError):
     """Error configuring the base."""
 
 
+class NetworkError(ProviderError):
```
mr-cal, line 29: Is this going to be used in a future PR?

mr-cal (dismissed):

facundobatista:
Mmm.... Now I doubt. Let's keep this branch, but close the PR, and see if we really spot this in the future.

PR #178, opened 2022-11-21
tests: refactor lxd launcher tests

mr-cal:

-----

The unit tests in tests/unit/lxd/test_launcher.py were designed to work with a single instance.  They use fixtures that do not distinguish which calls are made to which instance.

This is preventing me from testing https://github.com/canonical/craft-providers/pull/177, where I need the tests to distinguish between and instance and a base instance.

This PR refactors the fixtures and does a little cleanup.

(CRAFT-1478)

### DIFF
--- a/tests/unit/lxd/test_launcher.py
+++ b/tests/unit/lxd/test_launcher.py
@@ -16,7 +16,7 @@
 # Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #
 
-from unittest.mock import MagicMock, Mock, call, patch
+from unittest.mock import MagicMock, Mock, call
 
 import pytest
 
@@ -32,13 +32,10 @@ def mock_base_configuration():
 
 
 @pytest.fixture
-def mock_lxc():
-    with patch(
-        "craft_providers.lxd.launcher.LXC",
-        spec=lxd.LXC,
-    ) as mock_lxc:
-        mock_lxc.return_value.project_list.return_value = ["default", "test-project"]
-        yield mock_lxc.return_value
+def mock_lxc(mocker):
+    _mock_lxc = mocker.patch("craft_providers.lxd.launcher.LXC", spec=lxd.LXC)
+    _mock_lxc.return_value.project_list.return_value = ["default", "test-project"]
+    yield _mock_lxc.return_value
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


tests/unit/lxd/test_launcher.py
```diff
@@ -24,15 +25,15 @@
...
+    mock_base = Mock(spec=Base)
     mock_base.compatibility_tag = "mock-compat-tag-v100"
     mock_base.get_command_environment.return_value = {"foo": "bar"}
     yield mock_base
 
 
 @pytest.fixture
 def mock_lxc():
-    with mock.patch(
+    with patch(
```
sergiusens, line 36: maybe move this to use mocker?

cmatsuoka: +1 for the pytest mocker fixture

tests/unit/lxd/test_launcher.py
```diff
@@ -41,22 +42,35 @@ def mock_lxc():
...
         spec=lxd.LXDInstance,
-    ) as mock_instance:
-        mock_instance.return_value.name = "test-instance-$"
-        # the name has an invalid character, so the instance_name will be different
-        mock_instance.return_value.instance_name = "test-instance-fa2d407652a1c51f6019"
-        mock_instance.return_value.project = "test-project"
-        mock_instance.return_value.remote = "test-remote"
-        yield mock_instance
-
+        # single element list appears accidental but side_effect must be an interable.
```
cmatsuoka, line 65: Is this a typo for iterable?

tests/unit/lxd/test_launcher.py
```diff
@@ -41,22 +42,35 @@ def mock_lxc():
...
 @pytest.fixture
-def mock_lxd_instance():
-    with mock.patch(
+def fake_instance():
+    """Returns a fake LXD Instance"""
+    instance = MagicMock()
+    # the name has an invalid character to ensure the instance_name will be different
+    # so the two are not conflated in the unit tests
+    instance.name = "test-instance-$"
+    instance.instance_name = "test-instance-fa2d407652a1c51f6019"
```
cmatsuoka, line 51: I'm not familiar with the internal structure of instances but `instance.name` and `instance.instance_name` sound a bit confusing. Just mentioning, I don't think there's much we can do about that here in tests.

cmatsuoka (approved):
Thanks for updating. The changes are a bit hard to follow but the overall concept seems good. Pointed a possible typo in a comment.

tests/unit/lxd/test_launcher.py
mr-cal: I agree with you, it's a confusing subtlety that has arisen from craft-provider's evolving without breaking changes.

tests/unit/lxd/test_launcher.py
mr-cal: Yes, thanks.  Interesting that codespell doesn't catch it.

tests/unit/lxd/test_launcher.py
mr-cal: Added.

PR #182, opened 2022-12-06
bases: disable automatic snap refreshes

mr-cal:

-----
## Overview
Craft-providers manages the installation and versions of snaps inside the build environment, so automatic refreshes of snaps by snapd are disabled.

## Details
This functionality was imported from [snapcraft core18|20](https://github.com/snapcore/snapcraft/blob/31a3990c8f42eb27c326e42b7a79619996ab456d/snapcraft_legacy/internal/build_providers/_snap.py#L330).

It uses the command `snap set system refresh.hold=<1 day>`.  There is an [command](https://snapcraft.io/blog/hold-your-horses-i-mean-snaps-new-feature-lets-you-stop-snap-updates-for-as-long-as-you-need) in the snapd edge channel: `snap refresh --hold`.  Once this feature becomes stable, `craft-providers` can call this command once during setup.

(CRAFT-1518)


craft_providers/bases/buildd.py
```diff
@@ -421,6 +424,33 @@ def _disable_automatic_apt(
...
 
+    def _disable_and_wait_for_snap_refresh(
+        self, *, executor: Executor, deadline: Optional[float]
+    ) -> None:
+        """Disable automatic snap refreshes and wait for refreshes to complete.
+
+        Craft-providers manages the installation and versions of snaps inside the
+        build environment, so automatic refreshes of snaps by snapd are disabled.
+        """
+        # disable refresh for 1 day
```
cmatsuoka, line 435: Is 1 day long enough? (otoh if core 18-20 are using this it's probably better to stay consistent)

cmatsuoka (approved):
Sin this code was imported from core 18-20 I'm assuming most problems have already been solved in the previous implementation.

craft_providers/bases/buildd.py
mr-cal: If an craft lifecycle runs for more than 24 hours, then perhaps not.

I do agree that the consistency may be better.  There is no comment in core18|20 code for why 24 hours was chosen.

PR #185, opened 2022-12-07
docs: explanation on network issues

facundobatista:


mr-cal:
Regarding your offline comment about the double explanation header:
image

As the documentation grows, I imagine the explanation section will evolve into multiple pages.  Something like this:
image

I think you could rename your section from "Explanation" to "Network connectivity" or leave it as-is for now.  Either is fine with me.

mr-cal (approved):

facundobatista:
@mr-cal ack, thanks! We'll have this in mind for future docs improvements.

PR #188, opened 2022-12-09
lxd: delete and recreate invalid bases instances

mr-cal:

-----

To keep build environments clean, consistent, and up-to-date, any base instance older than 3 months (90 days) is deleted and recreated.

(CRAFT-1342)

### DIFF
--- a/craft_providers/lxd/launcher.py
+++ b/craft_providers/lxd/launcher.py
@@ -18,6 +18,7 @@
 """LXD Instance Provider."""
 
 import logging
+from datetime import datetime, timedelta
 from typing import Optional
 
 from craft_providers import Base, ProviderError, bases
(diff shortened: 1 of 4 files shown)


tests/integration/lxd/test_launcher.py
```diff
@@ -164,13 +166,84 @@ def test_launch_use_base_instance(get_base_instance, launch_args, instance_name)
...
+        assert instance.exists()
+        assert instance.is_running()
+    finally:
+        if instance.exists():
+            instance.delete()
+        if base_instance.exists():
+            base_instance.delete()
+
+
+@freeze_time(datetime.now() + timedelta(days=91))
```
lengau, line 182: Nice

tests/integration/lxd/test_launcher.py
```diff
@@ -164,13 +166,84 @@ def test_launch_use_base_instance(get_base_instance, launch_args, instance_name)
...
+        assert "'/base-instance': No such file or directory" in proc.stderr
+
+        # confirm new base instance does not have the expired base instance's
+        # fingerprint
+        base_instance.start()
+        proc = base_instance.execute_run(
+            ["stat", "/base-instance"], capture_output=True, text=True
+        )
+        assert proc.returncode == 1
+        assert "'/base-instance': No such file or directory" in proc.stderr
```
lengau, line 245: It feels like there are multiple tests in here. Can this be split into a few different, more specific tests?

Whether it can or not, this might also be a good use case for [pytest-check](https://github.com/okken/pytest-check), so we can complete the entire test and provide all the failures in one go (or groups of failures using the `any_failures()` function).

mr-cal: The purpose of this test is to follow this specific flow:
image

To setup for this, I first need to run through this flow:
image

This setup flow is tested in it's own separate test as well.  I added `asserts` to ensure the setup is successful.  Perhaps it could be moved to a fixture?  It's used by `test_launch_use_base_instance` and `test_launch_use_base_instance_expired`, so it would save some duplication and make the purpose of the test more clear.

Regarding the idea for pytest-check, I'm not sure. The function under test is sequential enough that once a single assert fails I won't benefit from seeing the subsequent failures.  And it may be more of a hindrance since bad calls to LXD can hang for a long time.

lengau: That's the same image twice, but it was enough for me to me comfortable with what you're doing and agree with your implementation.

I've found a couple of other places where we should be doing something that runs multiple asserts without failing fast, but it does sound this is very much not the place.

I'm happy with this.

lengau (approved):

tests/integration/lxd/test_launcher.py
lengau: Unfortunately you'll still need to get someone else on the team to approve since this newbie doesn't have write access yet :stuck_out_tongue:

mr-cal: Thanks!

Although those images are slightly different - I showed the path through by bolding the boxes and arrows in the flowchart.  The first image goes through two more steps in the middle of the flowchart than the second image.

lengau: Ohhhh I see now! The bolding didn't show up clearly on my phone. Sorry about that!

craft_providers/lxd/launcher.py
```diff
@@ -136,13 +137,58 @@ def _formulate_base_instance_name(
...
+        logger.warning("Instance does not have a 'Created' date.")
+        return False
+
+    # parse datetime
+    try:
+        creation_date = datetime.strptime(creation_date_raw, "%Y/%m/%d %H:%M %Z")
+    except ValueError as raised:
+        # if the date can't be parsed, consider it invalid
+        logger.warning(
+            "Could not parse instance's 'Created' date with error: %s", raised
```
facundobatista, line 178: Please also log with `%r` the `creation_date_raw`, just try to be as more explicit/precise as possible if a parse error happens...

craft_providers/lxd/launcher.py
```diff
@@ -136,13 +137,58 @@ def _formulate_base_instance_name(
     return "-".join(["base-instance", compatibility_tag, image_remote, image_name])
 
 
-# TODO: Implement this stubbed function (CRAFT-1341)
-# pylint: disable-next=unused-argument
-def _is_valid(instance: LXDInstance) -> bool:
-    """Check if a instance is valid (i.e. the base instance is not expired).
+def _is_valid(*, instance_name: str, project: str, remote: str, lxc: LXC) -> bool:
+    """Check if a instance is valid.
```
facundobatista, line 141: "an instance", right?

facundobatista (approved):
+1, but please check comments, thanks!!

craft_providers/lxd/launcher.py
mr-cal: Good idea

craft_providers/lxd/launcher.py
mr-cal: Yes!

PR #198, opened 2023-01-12
buildd: add try/except block for snap refresh commands

mr-cal:

-----

### DIFF
--- a/craft_providers/bases/buildd.py
+++ b/craft_providers/bases/buildd.py
@@ -496,18 +496,32 @@ class BuilddBase(Base):
 
         _check_deadline(deadline)
         # TODO: run `snap refresh --hold` once during setup (`--hold` is not yet stable)
-        executor.execute_run(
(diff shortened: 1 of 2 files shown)


sergiusens (approved):
thank you!

craft_providers/bases/buildd.py
```diff
@@ -496,18 +496,32 @@ def _disable_and_wait_for_snap_refresh(
...
-            capture_output=True,
-            check=True,
-        )
+        try:
+            executor.execute_run(
+                ["snap", "set", "system", f"refresh.hold={hold_time.isoformat()}Z"],
+                capture_output=True,
+                check=True,
+            )
+        except subprocess.CalledProcessError as error:
```
tigarmo, line 505: A bit late to the party (I see that this is already merged and that's fine) but maybe we should generalize this error handling a bit? Ideally we would always want to see the output of a failed command, right? Seems a bit annoying to have to remember to add try/except guards around every call to `execute_run()`
One idea would be to wrap the whole `setup()` call on a stop-gap try/catch

PR #199, opened 2023-01-20
lxd: set instance id map after launching or copying an instance

mr-cal:

-----
The id map for LXD instances needs to be set after an instance is launched from an image or after copying from a base instance.

(CRAFT-1577)

### DIFF
--- a/craft_providers/lxd/launcher.py
+++ b/craft_providers/lxd/launcher.py
@@ -267,7 +267,9 @@ def _set_id_map(
     :param remote: LXD remote to create instance on.
     :param uid: The uid to be mapped. If not supplied, the current user's uid is used.
     """
-    uid = uid if uid else os.getuid()
+    if uid is None:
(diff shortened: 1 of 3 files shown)


craft_providers/lxd/launcher.py
```diff
@@ -239,6 +244,39 @@ def _launch_existing_instance(
...
+
+    The instance needs to be stopped or restarted for the id map to take effect.
+
+    :param instance: LXD instance to set the idmap of
+    :param lxc: LXC client.
+    :param project: LXD project to create instance in.
+    :param remote: LXD remote to create instance on.
+    :param uid: The uid to be mapped. If not supplied, the current user's uid is used.
+    """
+    uid = uid if uid else os.getuid()
```
facundobatista, line 270: Better to be explicit, check `if uid is None`

facundobatista (approved):
Good work! Just check a minor comment.

PR #204, opened 2023-01-31
lxd: add default for _run_lxc() parameter `check=True`

mr-cal:

-----

No functional changes.

1. Add default for private method `run_lxd()`'s parameter `check=True` .  I did this because all calls to this method were setting `check=True`.
2. Reformat calls to `_run_lxc()` to improve readability


lengau (approved):
LGTM

PR #205, opened 2023-01-31
lxd: check id map when launching an existing instance

mr-cal:

-----

If an existing instance has a mismatched id map, either the instance will be auto-cleaned or a `BaseCompatibiltyError` will be raised.

(CRAFT-1576)

### DIFF
--- a/tests/integration/lxd/test_launcher.py
+++ b/tests/integration/lxd/test_launcher.py
@@ -523,6 +523,7 @@ def test_launch_os_incompatible(core20_instance):
         == "Incompatible base detected: Expected OS 'Ubuntu', found 'Fedora'."
     )
 
+    # TODO: split this into a separate test
     # when auto_clean is true, the instance will be deleted and recreated
     lxd.launch(
         name=core20_instance.name,
@@ -562,6 +563,7 @@ def test_launch_instance_config_incompatible(core20_instance):
         " Expected image compatibility tag 'buildd-base-v0', found 'invalid'."
     )
 
+    # TODO: split this into a separate test
     # when auto_clean is true, the instance will be deleted and recreated
     lxd.launch(
         name=core20_instance.name,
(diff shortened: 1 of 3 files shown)


tests/integration/lxd/test_launcher.py
```diff
@@ -573,3 +573,44 @@ def test_launch_instance_config_incompatible(core20_instance):
...
+            image_name="20.04",
+            image_remote="ubuntu",
+            map_user_uid=True,
+        )
+
+    assert exc_info.value.brief == (
+        "Incompatible base detected: "
+        "the instance's id map ('raw.idmap') is not configured as expected."
+    )
+
```
lengau, line 605: The bit below this and the bit above seem like two separate tests.

tests/unit/lxd/test_launcher.py
```diff
@@ -832,3 +922,73 @@ def test_set_id_map_all_options(fake_base_instance, mock_lxc, mocker):
...
+        (True, None, 5678, False),
+        # return False if id map is not expected and there is an id map
+        (False, 1234, None, False),
+        # return True if an id map is expected and the uid matches
+        (True, 1234, 1234, True),
+        # return False if an id map is expected and the uid does not match
+        (True, 1234, 5678, False),
+        # return True if an id map is expected and the uid matches the current users uid
+        (True, 101, None, True),
+    ],
```
lengau, line 943: This is fantastic, clear parametrization right here.

lengau (changes requested):
This looks great - the requested change is about the one test seeming like two, but I am also open to explanation about why it's not.

tests/unit/lxd/test_launcher.py
mr-cal: Thanks!

tests/integration/lxd/test_launcher.py
mr-cal: Yep.  The reason was `copypasta`, one of my common programming errors!

I split this apart and added a TODO for the tests I copied from.

lengau (approved):
Let's go!

Issue #190, opened 2022-12-15
Provide build base support for interim releases


coreycb:
This is a mirror of this bug for lpcraft: https://bugs.launchpad.net/lpcraft/+bug/1999694

The LXD provider only provides the ability to launch LTS releases. This prevents any testing of the current development (lunar) or any interim releases, such as kinetic. For lpcraft this is important because it is being used as a CI for packages, and all package updates must land in lunar and kinetic before they can land in any current LTS releases.

coreycb:
@mr-cal Thanks for triaging this. Is there anything I can do to help?

mr-cal:
Released with [craft-providers 1.8.0](https://pypi.org/project/craft-providers/1.8.0/)!

coreycb:
Thanks @mr-cal!!

PR #219, opened 2023-02-21
lxd: add support for interim Ubuntu releases

mr-cal:

-----
#### Overview
This PR adds support for interim and devel Ubuntu releases with LXD.

Broken into 6 separate commits to speed up review time.

#### Details

Previously, craft-providers maintained a list of valid LXD images with a simple enum.  Now, the `remotes.py` module has a new dataclass called `RemoteImage` and a dictionary of valid `RemoteImages` to track what images are supported.

It is designed in such a way that users can add custom remotes and we'll be able to add support for other bases (like a centos base) without breaking changes.

#### Usage

For applications using the `LxdProvider` class (snapcraft, charmcraft, and rockcraft), there are no changes in behavior or usage.  You can simply pass in the new release names:
```python
class BuilddBaseAlias(enum.Enum):
    XENIAL = "16.04"
    BIONIC = "18.04"
    FOCAL = "20.04"
    JAMMY = "22.04"
    KINETIC = "22.10"
    LUNAR = "23.04"
```

For other applications (lpcraft :eyes:), you'll need to use the replace the old `configure_buildd_image_remote()` function with the new `get_remote_image()` function.

Note - the old function `configure_build_image_remote()` still exists and its behavior is unchanged, except for a deprecation notice.

#### Example usage

```python
remote_image = lxd.get_remote_image(BuilddBaseAlias.KINETIC.value)

base_configuration = bases.BuilddBase(alias=BuildBaseAlias.KINETIC)

instance = lxd.launch(
    name="my-instance",
    base_configuration=base_configuration,
    image_name=remote_image.image_name,
    image_remote=remote_image.remote_name,
)
```

#### TODO
- [x] break into smaller commits

#### Sources

- [Provide build base support for interim releases #190](https://github.com/canonical/craft-providers/issues/190)
- [Support new series of ubuntu including interim releases ](https://bugs.launchpad.net/lpcraft/+bug/1999694)

(CRAFT-1601)

### DIFF
--- a/craft_providers/bases/buildd.py
+++ b/craft_providers/bases/buildd.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2021-2022 Canonical Ltd.
+# Copyright 2021-2023 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -144,6 +144,8 @@ class BuilddBaseAlias(enum.Enum):
     BIONIC = "18.04"
     FOCAL = "20.04"
     JAMMY = "22.04"
+    KINETIC = "22.10"
+    LUNAR = "23.04"
 
 
 class Snap(pydantic.BaseModel, extra=pydantic.Extra.forbid):
--- a/craft_providers/lxd/lxd_provider.py
+++ b/craft_providers/lxd/lxd_provider.py
@@ -19,14 +19,13 @@
 import contextlib
 import logging
 import pathlib
-import warnings
 from typing import Generator
 
 from craft_providers import Executor, Provider
 from craft_providers.base import Base
 from craft_providers.bases import BaseConfigurationError
 
-from .errors import LXDError
+from .errors import LXDError, LXDUnstableImageError
 from .installer import ensure_lxd_is_ready, install, is_installed
 from .launcher import launch
 from .lxc import LXC
@@ -99,6 +98,7 @@ class LXDProvider(Provider):
         base_configuration: Base,
         build_base: str,
         instance_name: str,
+        allow_unstable: bool = False,
     ) -> Generator[Executor, None, None]:
         """Configure and launch environment for specified base.
 
@@ -111,21 +111,20 @@ class LXDProvider(Provider):
         :param base_configuration: Base configuration to apply to instance.
         :param build_base: Base to build from.
         :param instance_name: Name of the instance to launch.
+        :param allow_unstable: If true, allow unstable images to be launched
 
         :raises LXDError: if instance cannot be configured and launched
         """
         image = get_remote_image(build_base)
         image.add_remote(lxc=self.lxc)
 
-        # we can't guarantee daily and devel images, so explicitly warn the user
-        if not image.is_stable:
-            warnings.warn(
-                message=(
-                    f"You are using an daily or devel image {image.image_name!r}"
-                    f" from remote {image.remote_name!r}. Devel or daily images are "
-                    "not guaranteed and are intended for experimental use only."
+        # only allow launching unstable images when opted-in with `allow_unstable`
+        if not image.is_stable and not allow_unstable:
+            raise LXDUnstableImageError(
+                brief=(
+                    f"Cannot launch an unstable image {image.image_name!r} from remote "
+                    f"{image.remote_name!r}"
                 ),
-                category=UserWarning,
             )
 
         try:
--- a/craft_providers/lxd/lxd_provider.py
+++ b/craft_providers/lxd/lxd_provider.py
@@ -115,7 +115,7 @@ class LXDProvider(Provider):
         :raises LXDError: if instance cannot be configured and launched
         """
         image = get_remote_image(build_base)
-        image.add_remote()
+        image.add_remote(lxc=self.lxc)
 
         # we can't guarantee daily and devel images, so explicitly warn the user
         if not image.is_stable:
--- a/craft_providers/lxd/lxd_provider.py
+++ b/craft_providers/lxd/lxd_provider.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright 2022 Canonical Ltd.
+# Copyright 2022-2023 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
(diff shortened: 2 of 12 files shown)


jugmac00:
From lpcraft side - it works like a charm! Thanks a lot!

craft_providers/lxd/remotes.py
```diff
@@ -15,50 +16,165 @@
...
+    is_stable: bool
+
+    def add_remote(self, lxc: LXC = LXC()) -> None:
+        """Add the LXD remote for an image.
+
+        If the remote already exists, it will not be re-added.
+
+        :param lxc: LXC client.
+        """
+        if self.remote_name in lxc.remote_list():
```
lengau, line 76: Theoretically this means the user could modify the remote and cause problems (images not available, wrong images coming, etc.). It might be worth checking the remote address in that case and providing a warning if it differs. (Or maybe that's too much of an edge case to care about considering that this is the same logic as previously existed.)

craft_providers/lxd/remotes.py
```diff
@@ -15,50 +16,165 @@
...
+    """Configure the default buildd image remote.
+
+    This is a deprecated function to maintain the existing API. It will be
+    removed with the release of craft-providers 2.0.
+
+    :param lxc: LXC client.
+
+    :returns: Name of remote to pass to launcher.
+    """
+    logger.warning(
```
lengau, line 173: I take it we don't use `warnings.warn` and `DeprecationWarning` here because it's filtered out by default?

tests/integration/lxd/test_lxd_provider.py
```diff
@@ -40,16 +42,26 @@ def test_create_environment(installed_lxd, instance_name):
...
-def test_launched_environment(installed_lxd, instance_name, tmp_path):
+@pytest.mark.parametrize(
+    "alias",
+    [
+        BuilddBaseAlias.BIONIC,
+        BuilddBaseAlias.FOCAL,
+        BuilddBaseAlias.JAMMY,
+        BuilddBaseAlias.KINETIC,
+        BuilddBaseAlias.LUNAR,
+    ],
```
lengau, line 53: Are we intentionally not testing on Xenial?

If we intend to test on all supported base aliases it may be better to do:

```suggestion
    "alias", list(BuilddBaseAlias),
```

so the tests automatically catch any new bases added.

craft_providers/lxd/remotes.py
mr-cal: That is what I surmised.  I am open to suggestions for a better approach to deprecations.

tests/integration/lxd/test_lxd_provider.py
mr-cal: `xenial` was conspicuously missing and untested in craft-providers (except one integration test).  It would raise an exception when trying to launch it using the Provider classes.

It's still available for LXD, but not for Multipass.  `lpcraft` still uses `xenial`.  

I'll add support for it and use your suggestion too!

craft_providers/lxd/remotes.py
mr-cal: You're right - it's a good check and an edge case.  I'll add a TODO for now

tests/integration/lxd/test_lxd_provider.py
mr-cal: After adding `xenial`, I'm seeing failures related to the version of `snapd` being too old.

craft_providers/lxd/remotes.py
```diff
@@ -58,6 +58,40 @@ class RemoteImage:
     remote_protocol: ProtocolType
     is_stable: bool
 
+    def add_remote(self, lxc: LXC = LXC()) -> None:
```
tigarmo, line 69: out of curiosity, what's the semantics of this default value? When would we use that?

craft_providers/lxd/lxd_provider.py
```diff
@@ -120,14 +114,26 @@ def launched_environment(
 
         :raises LXDError: if instance cannot be configured and launched
         """
-        image_remote = configure_buildd_image_remote()
+        image = get_remote_image(build_base)
+        image.add_remote()
+
+        # we can't guarantee daily and devel images, so explicitly warn the user
+        if not image.is_stable:
+            warnings.warn(
```
tigarmo, line 122: I thought `warnings.warn()` was meant more for devs, and not end-users

craft_providers/lxd/remotes.py
mr-cal: Updated to DeprecationWarning

craft_providers/lxd/lxd_provider.py
mr-cal: I believe @sergiusens requested this message be emitted with the `warnings` module so that it cannot be silenced or filtered out.

craft_providers/lxd/remotes.py
mr-cal: It's a great question and I don't have a good answer.

I've kept it because it was already there in the `configure_buildd_image_remote(lxc: LXC = LXC())` function.

I think this would be a good opportunity to remove the default value.  Done!

tests/integration/lxd/test_lxd_provider.py
lengau: Welp, now we know! Thank you!

craft_providers/lxd/lxd_provider.py
sergiusens: My thought process was that we were enabling the devel images early and its used would be something an application would enable, not a user through configuration. I would for example slightly change the API and add an allow_unstable=False param here and raise your warning as an exception.

craft_providers/bases/buildd.py
```diff
@@ -144,6 +144,8 @@ class BuilddBaseAlias(enum.Enum):
     BIONIC = "18.04"
     FOCAL = "20.04"
     JAMMY = "22.04"
+    KINETIC = "22.10"
+    LUNAR = "23.04"
```
sergiusens, line 148: There's a devel alias in the buildd-daily I would like to see here
image

mr-cal: This may be out of scope of this PR.  Adding the `DEVEL = "devel"` alias launches the image fine, but raises an error during setup because we are strictly matching `/etc/os-release` against the alias:

```
Incompatible base detected: Expected OS version 'devel', found '23.04'.
```

craft_providers/lxd/lxd_provider.py
mr-cal: As discussed offline, I changed this to an exception and made the `is_stable` property implicit.

sergiusens (approved):

jugmac00:
Quick heads-up - we faced a (short) [production regression](https://bugs.launchpad.net/lpcraft/+bug/2009954) after deploying these changes, as it looks like that after my review some more changes were made, especially the work around `add_remote`, which the example section does not cover.

As I have tried the previous state of that branch, my local lxc installation was already configured with the buildd remote (`craft-com.ubuntu.cloud-buildd`), but in production that failed.

Also, the changelog for 1.8.0 is missing, see https://github.com/canonical/craft-providers/issues/233 and https://github.com/canonical/craft-providers/blob/c6e624516dbb76ff60abfe9a2459bffe288d5e0e/docs/changelog.rst

Thanks to my colleague @lgp171188 we noticed that issue rather quickly, we rolled it back, and I was able to provide a fix for lpcraft.

Anyway, it works now, thanks a lot for your work!

P.S.: And looks like I need some more isolated test environment.

Issue #173, opened 2022-11-08
Interrupting an instance during setup may produce a bad instance


mr-cal:
## Issue
If an application is exited (`ctrl+c`) while an instance is being setup (i.e. during `bases.buildd.setup()`), then instance can get into a unknown state.  `craft-providers` does not track or check if an instance has been fully setup or not.  If the instance exists, it is assumed to be setup.

For example, `apt-update` may never run and subsequent attempts to install `stage-packages` will create an unhelpful error.

Another example is when the LXD or Multipass snap refreshes during container setup, which will likely cause `craft-providers` to raise an error and leave the container in an unknown state. (source @Guillaumebeuzeboc and @GAN-Canonical)

## Solution
We need to track the state of setup and warmup.  When an instance is successfully setup, we should write this in the instance configuration file.  

When restarting an existing instance, we should check the last setup time in the instance configuration file before choosing to `setup` or `warmup`.

`snapcraft core18|20` has solved this problem.  Perhaps the solution can be leveraged.

(CRAFT-343)

mr-cal:
Released with craft-providers 1.8.0

PR #223, opened 2023-02-24
bases(buildd): track if instance is properly setup

mr-cal:

-----

#### Overview
If craft-providers is interrupted while setting up an instance, the instance will be discarded on the next execution.

Broken into 5 discrete commits.

#### Details

The implementation is pretty straightforward, using the instance's instance configuration file:

```python
def setup():
    # set setup=False
    # set up instance...
    # set setup=True

def warmup():
    # check setup == True
    # warm up instance...
```

#### Background

Previously, craft-providers did not track if an instance was fully setup.  If the set up was interrupted, the next execution of craft-providers would assume the instance was setup.  This causes a whole suite of hard-to-troubleshoot failures (e.g. apt packages can't be installed, no internet connectivity, snapd not configured).

This is similar to what is done for snapcraft's [core18|20 build providers](https://github.com/snapcore/snapcraft/blob/5b0d370ed00a689dbc3754291605830f1a270329/snapcraft_legacy/internal/build_providers/_base_provider.py#L304).

#### Source
Resolves https://github.com/canonical/craft-providers/issues/173

(CRAFT-343)

### DIFF
--- a/craft_providers/bases/buildd.py
+++ b/craft_providers/bases/buildd.py
@@ -402,7 +402,7 @@ class BuilddBase(Base):
         self._setup_wait_for_system_ready(
             executor=executor, deadline=deadline, retry_wait=retry_wait
         )
-        self._setup_instance_config(executor=executor, deadline=deadline)
+        self._update_compatibility_tag(executor=executor, deadline=deadline)
         self._setup_hostname(executor=executor, deadline=deadline)
         self._setup_resolved(executor=executor, deadline=deadline)
         self._setup_networkd(executor=executor, deadline=deadline)
@@ -687,16 +687,6 @@ class BuilddBase(Base):
                 details=errors.details_from_called_process_error(error),
             ) from error
 
-    def _setup_instance_config(
-        self, *, executor: Executor, deadline: Optional[float]
-    ) -> None:
-        InstanceConfiguration.update(
-            executor=executor,
-            data={"compatibility_tag": self.compatibility_tag},
-            config_path=self.instance_config_path,
-        )
-        _check_deadline(deadline)
-
     def _setup_networkd(self, *, executor: Executor, deadline: Optional[float]) -> None:
         """Configure networkd and start it.
 
@@ -913,6 +903,17 @@ class BuilddBase(Base):
                 deadline, message="Timed out waiting for environment to be ready."
             )
             sleep(retry_wait)
+
+    def _update_compatibility_tag(
+        self, *, executor: Executor, deadline: Optional[float]
+    ) -> None:
+        """Update the compatibility_tag in the instance config."""
+        InstanceConfiguration.update(
+            executor=executor,
+            data={"compatibility_tag": self.compatibility_tag},
+            config_path=self.instance_config_path,
+        )
+        _check_deadline(deadline)
 
     def wait_until_ready(
         self,
--- a/craft_providers/bases/buildd.py
+++ b/craft_providers/bases/buildd.py
@@ -42,6 +42,8 @@ from .instance_config import InstanceConfiguration
 
 logger = logging.getLogger(__name__)
 
+# pylint: disable=too-many-lines
+
 
 def default_command_environment() -> Dict[str, Optional[str]]:
     """Provide default command environment dictionary.
@@ -268,6 +270,10 @@ class BuilddBase(Base):
     ) -> None:
         """Ensure instance configuration is compatible.
 
+        As long as the config is not incompatible (via a mismatched compatibility tag),
+        then assume the instance is compatible. This assumption is done is because the
+        config file may not exist or contain a tag while the set up is in progress.
+
         :raises BaseCompatibilityError: if instance is incompatible.
         :raises BaseConfigurationError: on other unexpected error.
         """
@@ -286,8 +292,8 @@ class BuilddBase(Base):
         except FileNotFoundError:
             return
 
-        # make the same assumption as above for empty configs
-        if config is None:
+        # make the same assumption as above for empty configs or configs without a tag
+        if config is None or config.compatibility_tag is None:
             return
 
         if config.compatibility_tag != self.compatibility_tag:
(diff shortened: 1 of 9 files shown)


craft_providers/bases/buildd.py
```diff
@@ -301,6 +303,55 @@ def _ensure_instance_config_compatible(
...
+            ) from error
+
+        if config is None:
+            raise BaseConfigurationError(
+                brief=brief,
+                details="Instance config is empty.",
+                resolution=resolution,
+            )
+
+        if not config.setup or config.setup is False:
```
tigarmo, line 346: What does this mean for existing instances (pre-`setup` attribute)? Are they invalidated anyway when craft-providers is updated?

tests/unit/bases/test_buildd.py
```diff
@@ -1023,6 +1059,79 @@ def test_ensure_config_compatible_empty_config_returns_none(fake_executor, mock_
...
+
+    base_config = buildd.BuilddBase(alias=buildd.BuilddBaseAlias.JAMMY)
+
+    with pytest.raises(BaseConfigurationError) as raised:
+        base_config._ensure_setup_completed(executor=fake_executor, deadline=None)
+
+    assert raised.value.brief == "Instance is not properly setup."
+    assert raised.value.details == details
+    assert (
+        raised.value.resolution == "Clean instance and retry the requested operation."
```
tigarmo, line 1098: So if I understand correctly, if the instance is not properly setup the user has to call `<application> clean` themselves right? Is this how it worked in core18/20? Have we thought about doing that for the user?

craft_providers/bases/buildd.py
mr-cal: Excellent question.  Right now, those existing instances would be considered misconfigured and the `BaseConfigurationError` would be raised.

I've been thinking this is an appropriate time to bump the base compatibility tag from `buildd-base-v0` to `buildd-base-v1`.  

This would prevent the existing instances would from being used.  New instances would be created instead.

tests/unit/bases/test_buildd.py
mr-cal: > So if I understand correctly, if the instance is not properly setup the user has to call <application> clean themselves right?
Correct.  

> Is this how it worked in core18/20?
`core18|20` auto-cleaned in this scenario.

When launching an instance, `craft-providers` will auto-clean or raise an exception, depending on the problem:
##### Reasons to autoclean
- wrong OS
- wrong LXD id mapping

##### Reasons to raise an Exception
- failure to install an apt package, snap
- failure to set hostname
- failure to setup network

I think you are right that this is more appropriate for an auto-clean rather than raising an Exception.

mr-cal: As discussed offline, I am now auto-cleaning when the instance is not fully setup, similar to core18|20.

craft_providers/bases/buildd.py
mr-cal: Bumped to `buildd-base-v1`

mr-cal:
Ready for review!

craft_providers/bases/buildd.py
```diff
@@ -268,6 +270,10 @@ def _ensure_instance_config_compatible(
     ) -> None:
         """Ensure instance configuration is compatible.
 
+        As long as the config is not incompatible (via a mismatched compatibility tag),
+        then assume the instance is compatible. This assumption is done is because the
```
tigarmo, line 274:
```suggestion
        then assume the instance is compatible. This assumption is done because the
```

craft_providers/bases/buildd.py
```diff
@@ -301,6 +307,41 @@ def _ensure_instance_config_compatible(
...
+            ) from error
+        except FileNotFoundError as error:
+            raise BaseCompatibilityError(
+                reason="failed to find instance config file",
+            ) from error
+
+        if config is None:
+            raise BaseCompatibilityError(reason="instance config is empty")
+
+        if not config.setup or config.setup is False:
```
tigarmo, line 340: this is a minor nit but isn't `if not config_setup` enough? In the case of `config.setup = False`, `not config.setup` will be True
I'm fine with leaving this as it is if you think it makes the intention clearer, though

tigarmo, line 340: So in the end what was the decision regarding auto-cleaning incomplete instances?

tests/integration/lxd/test_launcher.py
```diff
@@ -575,7 +592,55 @@ def test_launch_instance_config_incompatible_with_auto_clean(core20_instance):
...
+        file_mode="0644",
+    )
+
+    # when auto_clean is true, the instance will be deleted and recreated
+    lxd.launch(
+        name=core20_instance.name,
+        base_configuration=base_configuration,
+        image_name="20.04",
+        image_remote="ubuntu",
+        auto_clean=True,
```
tigarmo, line 605: ah I think this answers my question regarding the auto-clean

tests/unit/bases/test_buildd.py
```diff
@@ -42,11 +42,9 @@
 
 @pytest.fixture()
 def mock_load(mocker):
-    mocker.patch(
+    yield mocker.patch(
```
tigarmo, line 45: out of curiosity, why `yield` instead of `return`ing the patch object here?

tigarmo (changes requested):
nice work on the tests ;). Just some minor typos

craft_providers/bases/buildd.py
mr-cal: Thanks, this was a silly bit of logic.

craft_providers/bases/buildd.py
mr-cal: Thankss

tests/unit/bases/test_buildd.py
mr-cal: No good reason.  Most of the pytest fixtures yield instead of return, so I've been blindly following that practice.

Updated.

tests/integration/lxd/test_launcher.py
mr-cal: Bingo - the decision was to delete and recreate incomplete instances when `auto_clean=True`.

PR #228, opened 2023-03-03
bases: refactor BaseAlias for extra OS support

syu-w:
Change BaseAlias from str to tuple (OS, version)

-----

### DIFF
--- a/craft_providers/bases/__init__.py
+++ b/craft_providers/bases/__init__.py
@@ -17,14 +17,14 @@
 
 """Collection of bases used to configure build environments."""
 
+from .buildd import BaseAlias  # noqa: F401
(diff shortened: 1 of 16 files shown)


mr-cal:
This is an great start at adding CentOS support, but I think we need to take a different approach!

The `BuilddBase` class is Ubuntu-specific (well, other debian-based distros with `apt` would probably work). It would have been more clear if it had been named `UbuntuBase`.

For CentOS support, I think we need to create a new Base (`CentOSBase`) that also inherits from the `Base` class.

Some parts of the code in `BuilddBase` may be able to be moved into the parent class (or elsewhere), so it can be used in the CentOS base too.

syu-w:
Sounds good.

PR #231, opened 2023-03-08
bases(buildd): refactor `_ensure_os_compatible` into smaller functions

mr-cal:

-----
No functional changes.

The code to collect and parse `/etc/os-release` is split into a separate method (`_get_os_release()`) so it can be reused.

I will need to use this function very soon when updating the apt sources to support a `DEVEL` base!

(CRAFT-1636)

### DIFF
--- a/tests/unit/bases/test_buildd.py
+++ b/tests/unit/bases/test_buildd.py
@@ -561,12 +561,41 @@ def test_setup_timeout(fake_executor, fake_process, monkeypatch, mock_load):
     )
 
 
-def test_ensure_os_compatible_name_failure(fake_executor, fake_process):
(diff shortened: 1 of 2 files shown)


tests/unit/bases/test_buildd.py
```diff
@@ -561,54 +561,70 @@ def test_setup_timeout(fake_executor, fake_process, monkeypatch, mock_load):
     )
 
 
-def test_ensure_os_compatible_name_failure(fake_executor, fake_process):
+def test_get_os_release(fake_process, fake_executor):
```
mr-cal, line 564: I added an additional test and changed the order, so the diff is not helpful.

PR #232, opened 2023-03-09
bases(buildd): add new BuilddBaseAlias.DEVEL

mr-cal:

-----

- Add support for `BuilddBaseAlias.DEVEL`.
- Update `/etc/apt/sources.list` and `/etc/apt/sources.list.d/*.list` to point to the `devel` archives

(CRAFT-1636)

### DIFF
--- a/tests/integration/lxd/test_launcher.py
+++ b/tests/integration/lxd/test_launcher.py
@@ -116,18 +116,14 @@ def get_instance_and_base_instance(get_base_instance, instance_name):
             base_instance.delete()
 
 
-# exclude 23.04 (lunar) because it is a daily, not a supported release
-@pytest.mark.parametrize(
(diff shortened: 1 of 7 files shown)


tests/integration/lxd/test_launcher.py
```diff
@@ -116,18 +116,14 @@ def get_instance_and_base_instance(get_base_instance, instance_name):
             base_instance.delete()
 
 
-# exclude 23.04 (lunar) because it is a daily, not a supported release
-@pytest.mark.parametrize(
```
mr-cal, line 120: This test is about testing the launch command, not testing every single bases.

The integration tests in `tests/integration/lxd/test_remotes.py` is where we iterate through all the bases.

sergiusens (approved):
very nice and easy to follow

PR #238, opened 2023-03-10
provider: add `allow_unstable` to Provider base class

mr-cal:

-----

Adding a new parameter from `LXDProvider` to the `Provider` base class and the `Multipass` class.

Multipass is currently not using the parameter, but it will use it in a PR next week!

### DIFF
--- a/craft_providers/multipass/multipass_provider.py
+++ b/craft_providers/multipass/multipass_provider.py
@@ -103,6 +103,7 @@ class MultipassProvider(Provider):
         :param base_configuration: Base configuration to apply to instance.
         :param build_base: Base to build from.
         :param instance_name: Name of the instance to launch.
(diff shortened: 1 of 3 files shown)


craft_providers/multipass/multipass_provider.py
```diff
@@ -90,6 +90,7 @@ def launched_environment(
         base_configuration: base.Base,
         build_base: str,
         instance_name: str,
+        allow_unstable: bool = False,
```
cmatsuoka, line 93: We should add a `:param:` documentation for this argument in the method docstring.

mr-cal: I updated the docstrings!

cmatsuoka (approved):
Thanks!

PR #239, opened 2023-03-16
bases(buildd): update cloud.cfg to not reset apt's sources.list

mr-cal:

-----

When using a buildd base, `/etc/cloud/cloud.cfg` must be updated to prevent `/etc/apt/sources.list` from being reset when copying an instance.

This fixes an issue uncovered by https://github.com/snapcore/snapcraft/pull/4068

### DIFF
--- a/craft_providers/bases/buildd.py
+++ b/craft_providers/bases/buildd.py
@@ -1004,7 +1004,7 @@ class BuilddBase(Base):
         # get the current ubuntu codename
         os_release = self._get_os_release(executor=executor, deadline=deadline)
         version_codename = os_release.get("VERSION_CODENAME")
-        logger.debug("updating apt sources from %r to %r", version_codename, codename)
+        logger.debug("Updating apt sources from %r to %r.", version_codename, codename)
 
         # replace all occurrences of the codename in the `sources.list` file
         sed_command = ["sed", "-i", f"s/{version_codename}/{codename}/g"]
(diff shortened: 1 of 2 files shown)


craft_providers/bases/buildd.py
```diff
@@ -1015,6 +1016,20 @@ def _update_apt_sources(
                 details=errors.details_from_called_process_error(error),
             ) from error
 
+        # update cloud.cfg to prevent the sources.list file from being reset
+        try:
```
lengau, line 1020: Is there any time we want this to succeed if `cloud.cfg` doesn't exist?

craft_providers/bases/buildd.py
```diff
@@ -1015,6 +1016,20 @@ def _update_apt_sources(
...
+            _execute_run(
+                executor,
+                # 'aapt' is not a typo, the first 'a' is the sed command to append
+                # this is a shlex-compatible way to append to a file
+                ["sed", "-i", "$ aapt_preserve_sources_list: true", cloud_config],
```
lengau, line 1025: if cloud.cfg exists but is empty, this will exit with status 0 but not actually append to the file.

lengau (approved):
I've noted two potential edge cases. If we don't need to worry about them then I'm happy with this as is.

craft_providers/bases/buildd.py
mr-cal: Great idea.  `cloud.cfg` (from `cloud-init)`is fundamental to buildd images, so I decided to raise an error if it doesn't exist.

craft_providers/bases/buildd.py
mr-cal: Thanks, I added a check to make sure `cloud.cfg` exists and is not empty.

lengau (approved):

PR #240, opened 2023-03-17
lxd: set image description to instance full name

syu-w:
This allows other *craft to get info of the LXD instance versions.

### DIFF
--- a/craft_providers/lxd/launcher.py
+++ b/craft_providers/lxd/launcher.py
@@ -92,6 +92,15 @@ def _create_instance(
             project=project,
         )
 
+        # set the full instance name as image description
+        lxc.config_set(
+            instance_name=base_instance.instance_name,
+            key="image.description",
+            value=base_instance.name,
+            project=project,
+            remote=remote,
+        )
+
     # after creating the base instance, the id map can be set
     if map_user_uid:
         _set_id_map(instance=instance, lxc=lxc, project=project, remote=remote, uid=uid)
--- a/tests/integration/lxd/test_launcher.py
+++ b/tests/integration/lxd/test_launcher.py
@@ -16,6 +16,7 @@
 #
 
 import io
+import json
 import os
 import pathlib
 import subprocess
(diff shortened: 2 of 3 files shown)


craft_providers/lxd/launcher.py
```diff
@@ -92,6 +92,15 @@ def _create_instance(
             project=project,
         )
 
+    # set the full instance name as image description
+    lxc.config_set(
+        instance_name=instance.instance_name,
```
mr-cal, line 97: Should this be `base_instance.instance_name`?

craft_providers/lxd/launcher.py
```diff
@@ -92,6 +92,15 @@ def _create_instance(
...
+        key="image.description",
+        value=instance.name,
```
mr-cal, line 99: Should this be `base_instance.name`?

craft_providers/lxd/launcher.py
```diff
@@ -92,6 +92,15 @@ def _create_instance(
             project=project,
         )
 
+    # set the full instance name as image description
+    lxc.config_set(
```
mr-cal, line 96: Should this be indented (to be part of the if statement above)?  
I think this only needs to be written when the base instance is created.

mr-cal (changes requested):
Looking good!  

Can you add an integration test to confirm this command is working as expected?

You could probably sneak an assertion in [this test](https://github.com/canonical/craft-providers/blob/564b789112097f2b7c2f3115741276886f57e108/tests/integration/lxd/test_launcher.py#L165) or add another test.  Your choice.

craft_providers/lxd/launcher.py
syu-w: Yes, your are correct.

craft_providers/lxd/launcher.py
syu-w: Yes, this is suppose to be one more indent, but somehow I formatted it wrong.

mr-cal (approved):
Excellent work!  It's concise and has good comments and good tests.  Thank you!

PR #243, opened 2023-03-24
lint: remove 'module imported but unused' suppression

mr-cal:

-----

Flake8's [F401](https://www.flake8rules.com/rules/F401.html) warning (module imported but unused) does not need to be suppressed.


tigarmo (approved):
yesssss

PR #249, opened 2023-04-03
bases: add CentOS 7 support

syu-w:
Request Ubuntu 18.04 for cgroup v1 support

The image will pull CentOS image from LXD images servers.
Enable epel-release and centos-release-scl repos, and install backported python 3.8.
Python 3.8 will be the default python 3 interpreter.

All packages will be updated after bootstrapping finished.

CRAFT-1696

### DIFF
--- a/craft_providers/bases/__init__.py
+++ b/craft_providers/bases/__init__.py
@@ -25,6 +25,7 @@ from typing import Dict, Tuple, Type
 from craft_providers.errors import BaseCompatibilityError, BaseConfigurationError
 
 from ..base import Base
+from . import centos
 from . import ubuntu
 from . import ubuntu as buildd
 from .ubuntu import BuilddBase, BuilddBaseAlias
@@ -32,6 +33,8 @@ from .ubuntu import BuilddBase, BuilddBaseAlias
 sys.modules["craft_providers.bases.buildd"] = buildd
 
 __all__ = [
+    "ubuntu",
+    "centos",
     "BuilddBase",
     "BuilddBaseAlias",
     "BaseCompatibilityError",
@@ -46,6 +49,7 @@ BASE_NAME_TO_BASE_ALIAS: Dict[Tuple[str, str], Enum] = {
     ("ubuntu", "22.10"): ubuntu.BuilddBaseAlias.KINETIC,
     ("ubuntu", "23.04"): ubuntu.BuilddBaseAlias.LUNAR,
     ("ubuntu", "devel"): ubuntu.BuilddBaseAlias.DEVEL,
+    ("centos", "7"): centos.CentOSBaseAlias.SEVEN,
 }
 
 
--- a/craft_providers/bases/centos.py
+++ b/craft_providers/bases/centos.py
@@ -1,0 +1,836 @@
+#
+# Copyright 2021-2023 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+
+"""CentOS image(s)."""
+import enum
+import io
+import logging
+import pathlib
+import re
+import subprocess
+import sys
+import time
+from datetime import datetime, timedelta
+from time import sleep
+from typing import Dict, List, Optional, Type
+
+from pydantic import ValidationError
+
+from craft_providers import Base, Executor, errors
+from craft_providers.actions import snap_installer
+from craft_providers.actions.snap_installer import Snap, SnapInstallationError
+from craft_providers.errors import BaseCompatibilityError, BaseConfigurationError
+from craft_providers.util.os_release import parse_os_release
+
+from .instance_config import InstanceConfiguration
+
+logger = logging.getLogger(__name__)
+
+# pylint: disable=duplicate-code
+
+
+class CentOSBaseAlias(enum.Enum):
+    """Mappings for supported CentOS images."""
+
+    SEVEN = "7"
+
+
+class CentOSBase(Base):
+    """Support for CentOS images.
+
+    :cvar compatibility_tag: Tag/Version for variant of build configuration and
+        setup.  Any change to this version would indicate that prior [versioned]
+        instances are incompatible and must be cleaned.  As such, any new value
+        should be unique to old values (e.g. incrementing).  It is suggested to
+        extend this tag, not overwrite it, e.g.: compatibility_tag =
+        f"{appname}-{BuildBase.compatibility_tag}.{apprevision}" to ensure base
+        compatibility levels are maintained.
+    :cvar instance_config_path: Path to persistent environment configuration
+        used for compatibility checks (or other data).  Set to
+        /etc/craft-instance.conf, but may be overridden for application-specific
+        reasons.
+    :cvar instance_config_class: Class defining instance configuration.  May be
+        overridden with an application-specific subclass of InstanceConfiguration
+        to enable application-specific extensions.
+
+    :param alias: Base alias / version.
+    :param environment: Environment to set in /etc/environment.
+    :param hostname: Hostname to configure.
+    :param snaps: Optional list of snaps to install on the base image.
+    :param packages: Optional list of system packages to install on the base image.
+    """
+
+    compatibility_tag: str = f"centos-{Base.compatibility_tag}"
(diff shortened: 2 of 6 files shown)


tests/integration/lxd/test_remotes.py
mr-cal, line 1: Will you search through the test directories and add CentOS aliases where Buildd aliases are being iterated? (only where it's reasonable to test CentOS)

i.e.
```diff
diff --git a/tests/integration/lxd/test_remotes.py b/tests/integration/lxd/test_remotes.py
index 99bd45b..b382dc4 100644
--- a/tests/integration/lxd/test_remotes.py
+++ b/tests/integration/lxd/test_remotes.py
@@ -19,13 +19,13 @@ import subprocess

 import pytest

-from craft_providers import lxd
+from craft_providers import lxd, centos
 from craft_providers.bases import ubuntu


 # exclude XENIAL because it is not supported for LXD
 @pytest.mark.parametrize(
-    "alias", set(ubuntu.BuilddBaseAlias) - {ubuntu.BuilddBaseAlias.XENIAL}
+    "alias", set(ubuntu.BuilddBaseAlias) + set(centos.CentOSBaseAlias) - {ubuntu.BuilddBaseAlias.XENIAL}
 )
 def test_configure_and_launch_remote(instance_name, alias):
     """Verify remotes are configured and images can be launched."""
```

craft_providers/lxd/remotes.py
```diff
@@ -152,10 +153,16 @@ def add_remote(self, lxc: LXC) -> None:
         remote_address=DAILY_REMOTE_ADDRESS,
         remote_protocol=ProtocolType.SIMPLESTREAMS,
     ),
+    centos.CentOSBaseAlias.SEVEN.value: RemoteImage(
```
mr-cal, line 156: Feel free to refactor the `is_stable` property in this file if you want centos 7 to be considered stable.

Admittedly, I'm not happy with how I decided to implement remote names and remote urls in this file, but we're stuck with it for the short term.

craft_providers/lxd/lxd_provider.py
```diff
@@ -97,7 +97,7 @@ def launched_environment(
         project_name: str,
         project_path: pathlib.Path,
         base_configuration: Base,
-        build_base: str,
+        build_base: Tuple[str, str],
```
mr-cal, line 100: As far as I can tell, this is a transparent change for the craft applications, right?

Regardless, the Provider classes are labeled in-developer (and lpcraft isn't using them :see_no_evil:), so we have some wiggle room to make changes.

Can you update the base `Provider` class, the `MultipassProvider` class, and the docstrings?  The linters in the craft applications will catch the mismatched function signatures.

craft_providers/bases/centos.py
```diff
@@ -0,0 +1,950 @@
...
+        The state of the setup is tracked in the instance config. If the setup is
+        interrupted, the state can be checked so that a partially setup instance
+        can be discarded.
+
+        Setup may be called more than once in a given instance to refresh/update
+        the environment.
+
+        If timeout is specified, abort operation if time has been exceeded.
+
+        Guarantees provided by this setup:
```
mr-cal, line 351: Can you add anything noteworthy done during setup? Perhaps yum or the python setup.

mr-cal (changes requested):
Great work!  I am impressed with how you integrated this - it's a big effort.

My feedback is pretty minor, everything looks good overall.

craft_providers/bases/ubuntu.py
```diff
@@ -41,19 +41,20 @@
 
 logger = logging.getLogger(__name__)
 
+# pylint: disable=duplicate-code
```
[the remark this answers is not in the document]
syu-w: You mean the `duplicate-code`?

craft_providers/bases/centos.py
```diff
@@ -0,0 +1,950 @@
...
+                    )
+                except snap_installer.SnapInstallationError as error:
+                    raise BaseConfigurationError(
+                        brief=(
+                            f"failed to inject host's snap {snap.name!r} "
+                            "into target environment."
+                        )
+                    ) from error
+
+    def _setup_os_special_cmd(self, *, executor: Executor) -> None:
```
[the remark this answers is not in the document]
syu-w: CentOS has `network` enabled by default (At least working in my system). But that network configuration is not compatible with our current (`ubuntu` / `networkd`) configuration, and our all other bases use `networkd`. We can leave it alone, but I don't know if it will cause any issue.

craft_providers/lxd/lxd_provider.py
syu-w: This is transparent change unless something try to read the value and expect a str like "22.04".

craft_providers/bases/ubuntu.py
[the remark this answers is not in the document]
mr-cal: Whoops, I messed up that line.  I meant `# pylint: disable=duplicate-code, too-many-lines`.

But yes, pylint should be happy if you only have `# pylint: disable=duplicate-code` comment in 1 of the 2 instances of duplicated code.

craft_providers/bases/centos.py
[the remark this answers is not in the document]
mr-cal: If `network` worked by default for you, I vote we take the easiest path: use `network` and don't bother with switching to `systemd-networkd`.

If `network` doesn't work for another developer, then `yum` won't be able to download `systemd-networkd` and the developer will still be stuck with a broken instance.

If we run into network connectivity issues in the future (on launchpad, with proxies enabled, with docker installed, etc), we can explore some options, possibly:
1. Downloading `systemd-networkd` packages on the host and injecting them into the instance
2. Changing the `network` config settings to match `networkd`'s setting on the Ubuntu bases

tests/integration/lxd/test_remotes.py
syu-w: I am not sure if this one works in CI, as it needs 18.04. Let me try that.

craft_providers/lxd/remotes.py
syu-w: It probably not that stable. Those image from LXD are daily images.

craft_providers/bases/ubuntu.py
```diff
@@ -47,13 +47,13 @@
...
     """Mappings for supported buildd images."""
 
-    XENIAL = "16.04"
-    BIONIC = "18.04"
-    FOCAL = "20.04"
-    JAMMY = "22.04"
-    KINETIC = "22.10"
-    LUNAR = "23.04"
-    DEVEL = "devel"
+    XENIAL = ("ubuntu", "16.04")
```
mr-cal, line 50: Broken record - I'm still not convinced these tuples are adding value.

My concerns:
1. We can already deduce the base more pythonically using the type.  For example, the "ubuntu" and "centos" string aren't needed [here](https://github.com/canonical/craft-providers/pull/249/files#diff-2a629916bf70d9a5b8f644b9d2f37fe36447c8700d46990ab34619c72acd077aR67)
2. I'm not sure it's advantageous to the craft applications.  [This](https://github.com/canonical/charmcraft/pull/1065/files#diff-30a6f50c99bd977857f6dcacd6f4c935f66c39384725962288d623506e07398cR180) is still using type deduction, not the strings.  
3. And finally, although this eliminates the need for charmcraft's mapping `BASE_CHANNEL_TO_PROVIDER_BASE`, the other craft applications will still need a mapping, since they use different naming conventions.
4. The `is_base_available()` function in [charmcraft](https://github.com/canonical/craft-providers/pull/249/files#diff-2a629916bf70d9a5b8f644b9d2f37fe36447c8700d46990ab34619c72acd077aR57) uses the tuple, but that function really belongs in craft-providers. And I think there are better implementations for that validation, such as catching and re-raising the exception from your new `get_base_alias()` function.

craft_providers/bases/__init__.py
mr-cal, line 1: Can you unit-test the new functions in this class?

craft_providers/bases/ubuntu.py
syu-w: 1. and 2., because I see some place use the type, others use `base.value`. So if we can get every place use type, we don't need these tuples.

3. Maybe we can add theme together? like `core20`

4. OK

craft_providers/bases/__init__.py
syu-w: Sure.

craft_providers/bases/ubuntu.py
syu-w: And `_PROVIDER_BASE_TO_LXD_REMOTE_IMAGE` need to be changed if we want to accept `type` not `tuple` or `str` for mapping.

mr-cal: No problem - since it's a private member, feel free to change as needed!

craft_providers/lxd/remotes.py
```diff
@@ -75,10 +75,18 @@ def is_stable(self) -> bool:
...
+        if (
             self.remote_name == BUILDD_RELEASES_REMOTE_NAME
             and self.remote_address == BUILDD_RELEASES_REMOTE_ADDRESS
-        )
+        ):
+            # Ubuntu official buildd images
+            return True
+
+        if self.remote_name == "images":
+            # LXD daily images
```
mr-cal, line 86: Is this comment accurate?  Or is it about CentOS?

Either way, I think you could drop this if-statement since the function already returns false.

mr-cal (approved):
Excellent work!  It's been a long road and I think you nailed it!

PR #251, opened 2023-04-04
snap_installer: trim suffix from snap names

mr-cal:

-----

When installing a snap in an instance, do not use the snap name suffix (if present) when installing from the store, in temporary filenames, or in the instance config file.

This is a bug fix following up on CRAFT-1660, where the instance config files were not trimming the snap name suffix:

Previous `/etc/craft-instance.conf`:
```
compatibility_tag: snapcraft-buildd-base-v0.0
snaps:
  snapcraft:
    revision: '8931'
    source: host
  snapcraft_foo:
    revision: '8928'
    source: host
  snapcraft_bar:
    revision: '8931'
    source: host
```

New `/etc/craft-instance.conf`:
```
snaps:
  snapcraft:
    revision: '8931'
    source: host
```

### DIFF
--- a/craft_providers/actions/snap_installer.py
+++ b/craft_providers/actions/snap_installer.py
@@ -240,7 +240,8 @@ def _add_assertions_from_host(executor: Executor, snap_name: str) -> None:
     :param executor: Executor for target
     :param snap_name: Name of snap to inject
     """
-    target_assert_path = pathlib.Path(f"/tmp/{snap_name}.assert")
+    # trim the `_name` suffix, if present
+    target_assert_path = pathlib.Path(f"/tmp/{snap_name.split('_')[0]}.assert")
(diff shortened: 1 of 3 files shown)


lengau (approved):
Thank you for this! The code is clear and straightforward and the extra notes in docstrings are much appreciated.

I feel like this concept would be easy to accidentally break at a future point whilst still having the unit tests pass. Ideally, we'd have an end-to-end test covering this with an application, but since that's kinda off the table do you think an integration test to cover this would be feasible?

mr-cal:
I added to the existing integration tests to check the instance config file and added a new test to install a snap with a suffix from the store.

I couldn't find a good way to test injecting a snap installed with a suffix, since it requires rebooting the test instance.

lengau (approved):
Looks great! Thanks for humoring me

PR #261, opened 2023-05-03
bases: add almalinux 8/9 support

syu-w:
Install python 3.9 and building tools and git.


tests/unit/bases/test_almalinux.py
```diff
@@ -0,0 +1,1587 @@
...
+# pylint: disable=duplicate-code
+
+import subprocess
+from datetime import datetime
+from pathlib import Path
+from textwrap import dedent
+from unittest.mock import ANY, call, patch
+
+import pytest
+from logassert import Exact  # type: ignore
```
lengau, line 27: Can you make this ignore comment more specific?

tests/unit/bases/test_almalinux.py
```diff
@@ -0,0 +1,1587 @@
...
+from craft_providers.bases import almalinux
+from craft_providers.bases.instance_config import InstanceConfiguration
+from craft_providers.errors import (
+    BaseCompatibilityError,
+    BaseConfigurationError,
+    NetworkError,
+    details_from_called_process_error,
+)
+
+# pylint: disable=too-many-lines
```
lengau, line 40: Please put these together

tests/unit/bases/test_almalinux.py
syu-w: That's not possible, `# type: ignore`  is the recommended way to ignore missing typed library.

https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker

craft_providers/bases/__init__.py
lengau, line 1: Please update the copyright dates

craft_providers/bases/almalinux.py
```diff
@@ -0,0 +1,844 @@
...
+# pylint: disable=duplicate-code
+
+
+class AlmaLinuxBaseAlias(enum.Enum):
+    """Mappings for supported AlmaLinux images."""
+
+    NINE = "9"
+
+
+class AlmaLinuxBase(Base):
```
lengau, line 52: How different is this from the CentOS base?

syu-w: Probably about 95% are the same, so there's another PR to do a overall refactor.

lengau (approved):

PR #264, opened 2023-05-08
deps: pin urllib3 to v1.x

lengau:

-----

This is the same as craft-parts: https://github.com/canonical/craft-parts/pull/440


syu-w (approved):
Maybe `"urllib3 < 2.0.0"`?

lengau:
MacOS 10.15 has been removed from GitHub Actions, so these checks will remain queued until we move away from it.

Issue #169, opened 2022-10-28
Cannot upload files to `/root/` for Multipass


mr-cal:
Multipass instances cannot push files to `/root/`.  This is causing in issue in snapcraft, where we are attempting to push `/root/.bashrc` to the instance.

This is an [acknowledged behavior](https://github.com/canonical/multipass/issues/1433) in Multipass.

Possible solutions could be:
1. Push file to `/home/ubuntu/`, then execute `multipass exec -- sudo mv /home/ubuntu/... /root/...`
2. Updated ssh permission (see link above for details from Multipass team)

(CRAFT-1455)

PR #268, opened 2023-05-10
multipass(MultipassInstance): refactor push_file_io

mr-cal:
- [X] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

-----

## Overview

Refactor `push_file_io()`.  No functional changes.

## Details

I moved a block of code out of `push_file_io()` to `_create_temp_file()`, so I can reuse it in `push_file()` in a subsequent PR.

(CRAFT-1719)

PR #269, opened 2023-05-11
multipass(MultipassInstance): push files to any location

mr-cal:

-----

`push_file()` now follows the same design as `push_file_io()` where the file is first pushed to a temporary location and then moved to its final destination.

Blocked by https://github.com/canonical/craft-providers/pull/268
Resolves https://github.com/canonical/craft-providers/issues/169
(CRAFT-1719)


tests/unit/multipass/test_multipass_instance.py
```diff
@@ -573,6 +607,91 @@ def test_push_file_no_parent_directory(mock_multipass, instance, tmp_path):
...
+    error = subprocess.CalledProcessError(-1, ["mktemp"], "test stdout", "test stderr")
+
+    mock_multipass.exec.side_effect = [
+        # test call
+        mock.Mock(returncode=0),
+        # mktemp call
+        error,
+    ]
+
+    source = tmp_path / "src.txt"
```
sergiusens, line 621: you can probably create a small `test_file` fixture local to this test module to avoid this repetition

craft_providers/multipass/multipass_instance.py
```diff
@@ -393,10 +394,29 @@ def push_file(self, *, source: pathlib.Path, destination: pathlib.PurePath) -> N
                 f"Directory not found: {str(destination.parent.as_posix())!r}"
             )
 
-        self._multipass.transfer(
-            source=str(source),
-            destination=f"{self.name}:{destination.as_posix()}",
-        )
+        try:
+            tmp_file_path = self._create_temp_file()
```
sergiusens, line 398: this method assumes an ubuntu user; we should probably change our methods to have an execute_run argument to not sudo.

craft_providers/multipass/multipass_instance.py
```diff
@@ -372,15 +372,16 @@ def pull_file(self, *, source: pathlib.PurePath, destination: pathlib.Path) -> N
...
+        but the ownership is changed to the default user `ubuntu`.
 
         :param source: Host file to copy.
         :param destination: Target environment file path to copy to.  Parent
             directory (destination.parent) must exist.
 
         :raises FileNotFoundError: If source file or destination's parent
             directory does not exist.
-        :raises MultipassError: On unexpected error copying file.
+        :raises MultipassError: If the file cannot be pushed into the instance.
```
lengau, line 384: Do we want to provide other appropriate `OSError` subclasses, or should there never be other issues like permission errors?

tests/integration/multipass/test_multipass_instance.py
```diff
@@ -71,6 +71,40 @@ def test_push_file_io(reusable_instance, content, mode, user, group):
...
+@pytest.mark.parametrize(
+    "destination",
+    [
+        pathlib.Path("/tmp/push-file.txt"),
+        pathlib.Path("/home/ubuntu/push-file.txt"),
+        pathlib.Path("/root/push-file.txt"),
+    ],
+)
+@pytest.mark.parametrize("mode", ["644", "600", "755"])
+def test_push_file(destination, mode, reusable_instance, home_tmp_path):
```
lengau, line 83: Might be worth testing a couple of failures too. For example:

```python
@pytest.mark.parametrize(
    "destination",
    [
        pathlib.Path("/"),
        pathlib.Path("/non/existent/destination/file.txt"),
    ],
)
```

tests/unit/multipass/test_multipass_instance.py
mr-cal: Done!

craft_providers/multipass/multipass_instance.py
mr-cal: Added a new issue: https://github.com/canonical/craft-providers/issues/281

craft_providers/multipass/multipass_instance.py
mr-cal: I believe everything else will be caught and raised by Multipass.  For example, if there is a permission error on the source file, Multipass will fail and it will percolate up from `subprocess.CalledProcessError` to a `MultipassError`.

I did add an extra check to confirm the source is not a directory.

lengau: A part of me feels like I'd like to have more specific errors for what happens in the destination (e.g. an exception class that subclasses both `MultipassError` and `OSError` when the multipass issue is related to OS stuff), but I don't actually know whether that's useful or just me being naggy, so I'm gonna drop it for now :-)

lengau (approved):
Looks great! Thank you!

PR #283, opened 2023-05-30
bases: pre-installation of common packages

syu-w:
All images will be pre-installed with `build-essential` (or equivalent), `git`, and `python3` (with `devel`, `pip`, `venv`, `setuptools`, `wheel`)

### DIFF
--- a/craft_providers/bases/ubuntu.py
+++ b/craft_providers/bases/ubuntu.py
@@ -71,6 +71,7 @@ class BuilddBase(Base):
     :param hostname: Hostname to configure.
     :param snaps: Optional list of snaps to install on the base image.
     :param packages: Optional list of system packages to install on the base image.
+    :param use_default_packages: Optional bool to enable/disable default packages.
     """
 
     compatibility_tag: str = f"buildd-{Base.compatibility_tag}"
@@ -84,6 +85,7 @@ class BuilddBase(Base):
         hostname: str = "craft-buildd-instance",
         snaps: Optional[List[Snap]] = None,
         packages: Optional[List[str]] = None,
+        use_default_packages: bool = True,
     ):
         self.alias: BuilddBaseAlias = alias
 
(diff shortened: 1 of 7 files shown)


lengau:
Thinking purely in a `craft-providers` context, are these packages actually necessary?  I could imagine that the Python ones might be (if, for example, craft-providers runs Python code inside the provider), but these seem more application-specific otherwise.

syu-w:
Those packages (gcc, etc.) are actually pre-installed in the `buildd` image. So if only use that image, you don't need to add them. However, since we start support more general images, I think they should contain the same base components as buildd.

Also add them into base image will reduce the time spent on repeated downloads and installations those base packages.

craft_providers/bases/ubuntu.py
```diff
@@ -97,7 +97,7 @@ def __init__(
 
         self._set_hostname(hostname)
 
-        self._packages = ["apt-utils", "curl", "fuse", "udev"]
+        self._packages = ["apt-utils", "build-essential", "curl", "fuse", "udev"]
```
mr-cal, line 100: This is already installed on `buildd` images but installs extra packages on the `daily` interim images we are currently using?

syu-w: Yes, and apt will not try to install it again if it already installed.

mr-cal: Great!

mr-cal (approved):
I'm fine with all images matching `buildd`'s by having `build-essentials` and `python` pre-installed.  

Since Alex has concerns, will you wait to merge until you come to an agreement with him?

lengau:
Ideally I'd prefer for this to be the application's responsibility, since that also allows the app to do other preconfiguration that applies to all projects for that application (or even better — multiple snapshots for the same application depending on context). However, the perfect being the enemy of the good and all, now that I understand the underlying goal I'm going to withdraw my concerns since this still makes a whole lotta sense. Thanks for explaining! :-)

lengau (approved):

syu-w:
@mr-cal @lengau I added `use_default_packages` in case others want to override the default packages.

lengau (approved):

mr-cal (approved):

PR #301, opened 2023-06-09
lint: fix ruff warning PLC1901

mr-cal:

-----

Resolves:
- `PLC1901`: `string == ""` can be simplified to `not string` as an empty string is falsey

### DIFF
--- a/craft_providers/actions/snap_installer.py
+++ b/craft_providers/actions/snap_installer.py
@@ -73,7 +73,7 @@ class Snap(pydantic.BaseModel, extra=pydantic.Extra.forbid):
 
         :raises BaseConfigurationError: if channel is empty
         """
-        if channel == "":
+        if channel == "":  # noqa: PLC1901
             raise BaseConfigurationError(
                 brief="channel cannot be empty",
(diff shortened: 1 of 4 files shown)


craft_providers/actions/snap_installer.py
```diff
@@ -73,7 +73,7 @@ def validate_channel(cls, channel):
 
         :raises BaseConfigurationError: if channel is empty
         """
-        if channel == "":
+        if channel == "":  # noqa: PLC1901
```
lengau, line 76:
```suggestion
        if channel is not None and not channel:
```

I'm generally a fan of eliminating linter overrides, but I'm honestly not sure in this case which one I prefer.

lengau (approved):

craft_providers/actions/snap_installer.py
mr-cal: I'm 51/49 too - I think your's is slightly more explicit and clear.

mr-cal: Especially since the `# noqa: PLC1901` will make me stop and have to think about what is "wrong" with the if statement

PR #303, opened 2023-06-09
lint: fix ruff warnings SIM108, SIM112, SIM117, TRY300

mr-cal:

-----

Resolves: 
- `SIM108`: Use ternary operator instead of `if`-`else`-block
- `SIM112`: Use capitalized environment variable `HTTPS_PROXY` instead of `https_proxy`
- `SIM117`: Use a single `with` statement with multiple contexts instead of nested `with` statements
- `TRY300`: Consider moving this statement to an `else` block

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -228,6 +228,8 @@ ignore = [
     "D213",  # Multi-line docstring summary should start at the second line (reason: pep257 default)
     "D215",  # Section underline is over-indented (reason: pep257 default)
     "A003",  # Class attribute shadowing built-in (reason: Class attributes don't often get bare references)
+    "SIM117", # Use a single `with` statement with multiple contexts instead of nested `with` statements
+              # (reason: this creates long lines that get wrapped and reduces readability)
 
     # Ignored due to common usage in current code
     "TRY003",  # Avoid specifying long messages outside the exception class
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -235,12 +235,8 @@ ignore = [
     "S324", # Probable use of insecure hash functions in `hashlib`
     "S506", # Probable use of unsafe loader `BaseLoader` with `yaml.load`. Allows instantiation of arbitrary "object"s. Consider `yaml.safe_load`.
     "SIM102", # Use a single `if` statement instead of nested `if` statements
-    "SIM108", # Use ternary operator instead of `if`-`else`-block
-    "SIM112", # Use capitalized environment variable `HTTPS_PROXY` instead of `https_proxy`
     "SIM114", # Combine `if` branches using logical `or` operator
-    "SIM117", # Use a single `with` statement with multiple contexts instead of nested `with` statements
```
lengau, line 241: I like this in theory, but in practice the resulting code here feels a bit weird to me. idk wdyt?

tests/unit/multipass/test_multipass.py
```diff
@@ -728,9 +729,10 @@ def test_version_decode_error(fake_process):
         ["multipass", "version"], stdout=b"multipass 1.6.2\xd0"
     )
 
-    with mock.patch("locale.getpreferredencoding", return_value="ascii"):
-        with pytest.raises(MultipassError) as exc_info:
-            Multipass().version()
+    with mock.patch("locale.getpreferredencoding", return_value="ascii"), pytest.raises(
```
lengau, line 732: OK I especially don't like SIM117 in tests here, since the two context managers are conceptually quite distinct in their purposes (unlike in the production code where they're related).

Having seen this I think I have a preference for ignoring SIM117 for tests at least.

lengau (approved):
In theory, theory and practice are the same thing. But in practice, sometimes theory is ugly.

tests/unit/multipass/test_multipass.py
mr-cal: From our offline conversation, we agreed that combining `with` statements and having an 88 character line limit makes the code less readable.  

I undid the SIM117 changes, ignored the rule, and pushed the change to starbase. https://github.com/canonical/starbase/pull/90

PR #305, opened 2023-06-10
base: clean up retry-until-timeout logic

lengau:

-----


tests/unit/test_util.py
mr-cal, line 1: Should this be `tests/unit/util/test_retry.py`?

tests/unit/test_util.py
```diff
@@ -0,0 +1,53 @@
...
+    return request.param * 0.001
+
+
+@pytest.fixture(params=range(5), ids=[f"wait_{i}us" for i in range(5)])
+def retry_wait(request):
+    """A parametrized fixture of times to wait before retrying, in microseconds"""
+    return request.param * 0.000_001
+
+
+def test_retry_until_timeout_success(failure_count, retry_wait):
```
mr-cal, line 35: Do we need to test a scenario where `retry_wait` > `timeout`?

Asking because that was the original bug that began this.

mr-cal (approved):
Nice work!

tests/unit/test_util.py
lengau: Yep! I missed the existence of the `util` package at first while I was writing everything and forgot to move the tests over

craft_providers/util/retry.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    deadline = time.monotonic() + timeout
+
+    while time.monotonic() < deadline:
+        try:
+            return func(deadline - time.monotonic())
+        except Exception:
+            if time.monotonic() < deadline - retry_wait:
+                time.sleep(retry_wait)
+    try:
```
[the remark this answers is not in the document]
lengau: We could exclude it, but all the use cases we had included converting the error anyway, so I figured this could be responsible for converting the error too.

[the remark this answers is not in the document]
mr-cal: Oh, nevermind.  In my head, I read it as:
```python
try:
  return func(retry_wait)
except Exception as exc:
  raise exc
```

mr-cal:
The `test_retry` unit tests are unreliable on the windows runners.  

It looks like the runners are slow and can't get through the retry logic within the 1-5ms time window.

PR #309, opened 2023-06-13
tests: use fixture for base_configuration

mr-cal:

-----

Minor refactor. Prerequisite for CRAFT-1767.


mr-cal:
Canceling this workflow so I can get some other PRs tested first...

lengau (approved):
This is really nice! And a great side-effect is that if we want to do this with other bases in future we can just parametrise the `base_configuration` fixture and get all the tests for free. (Not right now because... well, y'know, CI times)

Issue #311, opened 2023-06-13
Add `mxschmitt/action-tmate` to workflow


mr-cal:
https://github.com/marketplace/actions/debugging-with-tmate provides ssh access into runners.

It is designed to live permanently in a workflow.  It can be triggered by enabling `debug` when running an action, so no code changes or test branches are needed.

If it is useful, we could upstream the workflow changes to starbase.

PR #313, opened 2023-06-14
ci: add permanent ssh access to github runners

mr-cal:

-----

Add ssh access to runners in the CI workflow using [action-tmate](https://github.com/marketplace/actions/debugging-with-tmate).

This is to live permanently in the workflow. It can be triggered by enabling debug when running an action, so no code changes or test branches are needed.

If it is useful, I can upstream the workflow changes to `starbase`.

This PR requires a leap of faith -  we need to merge the PR and then confirm it works as expected.  It worked great on my own fork, so I expect it will work here.

Resolves https://github.com/canonical/craft-providers/issues/311

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -8,6 +8,15 @@ on:
       - "release/*"
   pull_request:
 
+  # Allows triggering the workflow manually from the Actions tab
+  workflow_dispatch:
+    inputs:
+      enable_ssh_access:
+        type: boolean
+        description: 'Enable ssh access'
+        required: false
+        default: false
+
 jobs:
   linters:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yaml
```diff
@@ -38,6 +47,9 @@ jobs:
           echo "::group::Wait for snap to complete"
           snap watch --last=install
           echo "::endgroup::"
+      - name: Enable ssh access
+        uses: mxschmitt/action-tmate@v3
+        if: ${{ inputs.enable_ssh_access }}
```
lengau, line 52: Should we set this with `limit-access-to-actor: true`? It concerns me a bit that the default is to allow anyone to connect to the runner.

lengau (changes requested):

.github/workflows/tests.yaml
mr-cal: Great idea, I added that flag and updated the documentation.  

I'll test that it is enforced after this gets merged.

lengau (approved):

PR #314, opened 2023-06-14
lxd: clear hostname of base instance

mr-cal:

-----

The hostname is cleared before creating a LXD base instance from an instance. The hostname is set when creating a new instance from the base instance.

Depends on https://github.com/canonical/craft-providers/pull/305 (which adds some unit test fixtures for the `Base` class), so only the last commit is relevant here.

Source: 
- https://github.com/snapcore/snapcraft/issues/4181

(CRAFT-1767)

### DIFF
--- a/craft_providers/base.py
+++ b/craft_providers/base.py
@@ -332,6 +332,20 @@ class Base(ABC):
             error=error,
         )
 
+    def clear_hostname(self, executor: Executor) -> None:
+        """Clear hostname by removing /etc/hostname."""
+        try:
+            self._execute_run(
+                ["rm", "-f", "/etc/hostname"],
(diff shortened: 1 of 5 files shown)


lengau (approved):
:-D

mr-cal:
I'm having some trouble with LXD instances recognizing craft-providers' manual edit of `/etc/hostname`.  

I think a better solution is:
1. Stop editing `/etc/hostname`
2. Rely on LXD to set the hostname to the instance name

If we really want to manually set the hostname to something besides the instance name, then we should try using `hostnamectl`

mr-cal:
Closing in favor of https://github.com/canonical/craft-providers/pull/337

Issue #329, opened 2023-06-22
Flaky tests 


lengau:
This is a collector for logging which CI tests are flaky. Because of the nature of craft-providers, CI can have a lot of different things go wrong that are beyond our control, so in some cases we just need to mark tests as flaky and allow re-runs. It's better to re-run individual tests when possible rather than completely re-running the entire test suite, especially given how long the entire integration test suite takes.

Currently flaky:

* [tests/integration/lxd/test_launcher.py::test_launch_use_base_instance_expired](https://github.com/canonical/craft-providers/actions/runs/5347071930/jobs/9695009814#step:6:60)
* `tests/unit/util/test_retry.py` on Windows
* [lxd/test_launcher.py::test_launch_use_base_instance_expired](https://github.com/canonical/craft-providers/actions/runs/5422878603/jobs/9860082854?pr=345#step:7:48)

PR #331, opened 2023-06-22
base: retry retrieving /etc/os-release

lengau:

-----

This occasionally fails if the CPU is busy (most noticeable in CI). Retrying should reduce the chance of that.

### DIFF
--- a/craft_providers/base.py
+++ b/craft_providers/base.py
@@ -231,25 +231,39 @@ class Base(ABC):
 
         :returns: Dictionary of key-mappings found in os-release.
         """
-        try:
-            # Replace encoding errors if it somehow occurs with utf-8. This
-            # doesn't need to be perfect for checking compatibility.
-            proc = executor.execute_run(
-                command=["cat", "/etc/os-release"],
-                capture_output=True,
-                check=True,
-                text=True,
-                encoding="utf-8",
-                errors="replace",
-                timeout=self._timeout_simple,
-            )
-        except subprocess.CalledProcessError as error:
-            raise BaseConfigurationError(
-                brief="Failed to read /etc/os-release.",
-                details=details_from_called_process_error(error),
-            ) from error
 
-        return parse_os_release(proc.stdout)
+        # `lxc exec` will occasionally print an empty string when the CPU is
+        # busy. The retry logic here decreases the chance of that.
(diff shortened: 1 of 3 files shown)


syu-w:
How the CI timeout a `cat` 😑

lengau:
@syu-w it's not timing out on `cat` — `lxc exec` occasionally sends an empty string to stdout when the machine is very busy. My hypothesis is that there's a race condition in lxc related to the pipes, but I haven't been able to replicate it on my own machine (only in CI).

syu-w:
So it send an empty string and exit? Sounds like it got killed?

craft_providers/util/retry.py
```diff
@@ -31,28 +31,32 @@ def retry_until_timeout(
...
     :param retry_wait: The length of time (in seconds) before retrying.
     :param func: The callable. May only take a timeout parameter.
         Must raise an exception on failure, may return anything on success.
-    :param error: Exception to raise on timeout
+    :param error: Exception to raise on timeout or None to pass the error unchanged.
     :returns: The result of the function
     :raises: the passed error from the last exception
     """
     deadline = time.monotonic() + timeout
+    print(f"deadline {deadline=}")
```
mr-cal, line 47: These `print`s are temporary / will be removed before merging?

mr-cal (approved):

craft_providers/util/retry.py
lengau: Ope, yep!

lengau:
@syu-w I don't think so, because it has the correct return code on exit (the same return code as the process we exec). I think more likely there's a race condition related to flushing the stdout and stderr streams before exit, and something about the Github Actions environment makes the erroneous outcome far more likely than I've been able to reproduce on my machine in any reliable manner. 

I've created https://github.com/lxc/lxd/issues/11890 to see if the LXD team has any idea, but honestly it seems like it'd be very hard to find, since I can't even reliably reproduce it.

syu-w:
Then there is possible other `exec` will also had this issues.

lengau:
Agreed, but this is the one that I've seen enough to make it into my flaky tests dashboard.  (See #329 ). And given that changing the `exec` function to retry could be problematic I don't want to change it there.

PR #337, opened 2023-06-26
bases: use default hostnames

mr-cal:
- [X] Have you followed the guidelines for contributing?

-----

Let LXD and Multipass set the instance's hostname to the name. LXD names are compatible with hostname naming conventions. Multipass names are compatible but do not have a maximum length. Multipass truncates the hostname to the first 64 characters of the name.

Note - this is a hotfix for the `1.10` branch for snapcraft `7.4`.

This required 2 cherry-picks from `main`.

(CRAFT-1767)

### DIFF
--- a/craft_providers/bases/buildd.py
+++ b/craft_providers/bases/buildd.py
@@ -21,7 +21,6 @@ import io
 import logging
 import os
 import pathlib
-import re
 import subprocess
 import sys
 import time
@@ -202,7 +201,7 @@ class BuilddBase(Base):
 
     :param alias: Base alias / version.
     :param environment: Environment to set in /etc/environment.
-    :param hostname: Hostname to configure.
+    :param hostname: Deprecated parameter, not used.
     :param snaps: Optional list of snaps to install on the base image.
     :param packages: Optional list of system packages to install on the base image.
     """
@@ -217,7 +216,7 @@ class BuilddBase(Base):
         alias: BuilddBaseAlias,
         compatibility_tag: Optional[str] = None,
         environment: Optional[Dict[str, Optional[str]]] = None,
-        hostname: str = "craft-buildd-instance",
+        hostname: Optional[str] = None,
         snaps: Optional[List[Snap]] = None,
         packages: Optional[List[str]] = None,
     ):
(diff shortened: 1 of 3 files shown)


mr-cal:
Craft-providers no longer touches the hostname and trusts LXD and Multipass to manage the hostname throughout an instance's lifecycle (whether an instance is launched, stopped and started, copied, etc).

Because it's entirely out of craft-providers' purview, craft-providers does not need integration tests for the hostname.

What do you think, @lengau?  I don't mind adding tests if you see value but since the integration tests are slow, I'm trying to be thoughtful about what is important to test.

.github/workflows/tests.yaml
lengau, line 1: The changes to this file seem independent of the rest of the PR?

.github/workflows/tests.yaml
```diff
@@ -83,8 +82,10 @@ jobs:
         run: |
           brew update
           brew install multipass
-          multipass version
           sleep 20
+          multipass set local.driver=qemu
```
lengau, line 86: I thought this was fixed with the latest multipass?

setup.py
lengau, line 1: Like with `tests.yaml`, this seems unrelated to this PR other than the build potentially breaking if we don't include it.

Could you move this and `tests.yaml` to a separate PR that we can merge first?

lengau (changes requested):
I'm good with nuking the hostname integration tests. If we're not managing the hostname it's not our job to bother with them. 

What was the reason for setting the hostnames in the first place? Does something depend on the hostnames having specific values?

(Changes requested are a separate PR for the metadata changes)

setup.py
mr-cal: Sure, I pulled the prereqs to https://github.com/canonical/craft-providers/pull/338

.github/workflows/tests.yaml
mr-cal: Ah yes, thanks for the reminder

mr-cal:
> What was the reason for setting the hostnames in the first place? Does something depend on the hostnames having specific values?

I'm not sure and I don't think anybody on the team knows.  I don't believe any of the craft-applications depend on the hostnames, as no one has raised a problem about the [hostname mismatch bug](https://github.com/snapcore/snapcraft/issues/4181), other than it being distracting.

.github/workflows/tests.yaml
mr-cal: Well, apparently it's not!  I wonder why this is working on `main`...

https://github.com/canonical/multipass/releases/tag/v1.12.0

https://github.com/canonical/craft-providers/actions/runs/5392910938/jobs/9791898806?pr=338#step:5:1187

lengau: pakistani_cricket_fan_disappointed.gif

mr-cal:
Running in circles.  Hostnames are managed by `cloud-init`, which exists on images from https://cloud-images.ubuntu.com but does not exist on images from https://cloud-images.ubuntu.com/buildd

So all my previous conclusions are wrong.  Craft-providers needs to manually set the hostname for buildd images.  
And clearing the hostname (deleting /etc/hostname) is a bad idea.  It is causing unpredictable behavior with setting the hostname.

Moving this to a draft.

mr-cal:
Superceded by https://github.com/canonical/craft-providers/pull/343

PR #343, opened 2023-06-29
lxd: set hostname after copying from a base instance

mr-cal:

-----

When copying from a base instance, set the hostname of the instance.

This is a replacement for #337, where I thought that hostnames would be self-managed.  That is only the case for images with `cloud-init` configured and buildd images do not have `cloud-init` :sob: 

I also decided not to "clear" (remove) the base instance hostname.  It led to unpredictable behavior the next time I tried to set it.

(CRAFT-1767)

### DIFF
--- a/craft_providers/lxd/launcher.py
+++ b/craft_providers/lxd/launcher.py
@@ -615,5 +615,10 @@ def launch(
     # instance is now ready to be started and warmed up
     logger.warning("Starting instance.")
     instance.start()
+
+    # change the hostname from the base instance's hostname
+    # pylint: disable-next=protected-access
+    base_configuration._setup_hostname(executor=instance, deadline=None)  # type: ignore
+
     base_configuration.warmup(executor=instance)
     return instance
--- a/tests/integration/lxd/test_launcher.py
+++ b/tests/integration/lxd/test_launcher.py
@@ -218,7 +218,7 @@ def test_launch_use_base_instance(get_instance_and_base_instance, instance_name)
 
     # collect the hostname of the instance
     instance_hostname = (
-        instance.execute_run(["cat", "/etc/hostname"], check=True, capture_output=True)
+        instance.execute_run(["hostname"], check=True, capture_output=True)
         .stdout.decode()
         .rstrip("\n")
     )
--- a/tests/integration/lxd/test_launcher.py
+++ b/tests/integration/lxd/test_launcher.py
@@ -216,6 +216,15 @@ def test_launch_use_base_instance(get_instance_and_base_instance, instance_name)
     assert instance.exists()
     assert instance.is_running()
 
+    # collect the hostname of the instance
+    instance_hostname = (
+        instance.execute_run(["cat", "/etc/hostname"], check=True, capture_output=True)
+        .stdout.decode()
+        .rstrip("\n")
+    )
+
+    assert instance_hostname == base_configuration.hostname
+
 
 def test_launch_create_base_instance_with_correct_image_description(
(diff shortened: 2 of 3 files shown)


craft_providers/lxd/launcher.py
```diff
@@ -615,5 +615,10 @@ def launch(
     # instance is now ready to be started and warmed up
     logger.warning("Starting instance.")
     instance.start()
+
+    # change the hostname from the base instance's hostname
+    # pylint: disable-next=protected-access
+    base_configuration._setup_hostname(executor=instance)  # type: ignore
```
mr-cal, line 621: This ignored type can be removed once this is merged into the main branch.

craft_providers/lxd/launcher.py
```diff
@@ -615,5 +615,10 @@ def launch(
     # instance is now ready to be started and warmed up
     logger.warning("Starting instance.")
     instance.start()
+
+    # change the hostname from the base instance's hostname
+    # pylint: disable-next=protected-access
+    base_configuration._setup_hostname(executor=instance, deadline=None)  # type: ignore
```
mr-cal, line 621: This will look cleaner on the main branch (no linter disabling, no need to pass a deadline)

mr-cal:
I'm not going to wait on the multipass tests, since this is a LXD only change.

tests/integration/lxd/test_launcher.py
```diff
@@ -216,6 +216,15 @@ def test_launch_use_base_instance(get_instance_and_base_instance, instance_name)
     assert instance.exists()
     assert instance.is_running()
 
+    # collect the hostname of the instance
```
mr-cal, line 219: Ideally, I would add another test that would set and verify hostnames before and after copying them, but I'm running out of time for this sprint.

tests/unit/lxd/test_launcher.py
```diff
@@ -26,12 +26,12 @@
 from freezegun import freeze_time
 from logassert import Exact  # type: ignore
 
-from craft_providers import Base, ProviderError, bases, lxd
+from craft_providers import ProviderError, bases, lxd
 
 
 @pytest.fixture
 def mock_base_configuration():
-    mock_base = Mock(spec=Base)
```
lengau, line 34: I'm hesitant about this because it could result in us missing changes we need to make to other bases. What's the reasoning behind it, and is there a way we can still spec it on Base instead (or parametrize the fixture to use each concrete base)?

mr-cal: I did this because `_setup_hostname()` method isn't in the parent `Base` class.

Note - this is on `1.10` where the only base is `buildd`.

Because of Sheng's base refactoring in `1.14`, I don't have to do this "hack" on `main` (meaning I won't bring in this change when I merge this hotfix branch into `main`).

tests/integration/lxd/test_launcher.py
```diff
@@ -216,6 +216,15 @@ def test_launch_use_base_instance(get_instance_and_base_instance, instance_name)
...
+    instance_hostname = (
+        instance.execute_run(["cat", "/etc/hostname"], check=True, capture_output=True)
```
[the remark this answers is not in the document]
mr-cal: Good question, I changed it to `hostname`.

tests/unit/lxd/test_launcher.py
lengau: Ahh, got it. I'm good with that then :-)

lengau (approved):
Looks good!

PR #357, opened 2023-07-13
lxc: recheck when creating project failed

syu-w:
To avoid race when multiple caller try to create same project.

CRAFT-1849

### DIFF
--- a/craft_providers/lxd/lxc.py
+++ b/craft_providers/lxd/lxc.py
@@ -793,10 +793,19 @@ class LXC:
         try:
             self._run_lxc(command, capture_output=True, project=project)
         except subprocess.CalledProcessError as error:
-            raise LXDError(
-                brief=f"Failed to create project {project!r}.",
-                details=errors.details_from_called_process_error(error),
-            ) from error
+            # handle the race condition where two processes check and
+            # create the same project at the same time
+            if project in self.project_list(remote=remote):
+                logger.debug(
+                    "Remote %s is present on second check, ignoring exception %s.",
+                    project,
+                    str(error),
+                )
+            else:
+                raise LXDError(
+                    brief=f"Failed to create project {project!r}.",
+                    details=errors.details_from_called_process_error(error),
+                ) from error
 
     def project_delete(self, *, project: str, remote: str = "local") -> None:
         """Delete project, if it exists.
--- a/tests/unit/lxd/test_lxc.py
+++ b/tests/unit/lxd/test_lxc.py
@@ -1485,6 +1485,17 @@ def test_project_create_error(fake_process):
         ],
         returncode=1,
     )
+    fake_process.register_subprocess(
+        [
+            "lxc",
+            "project",
+            "list",
+            "test-remote:",
+            "--format=yaml",
+        ],
+        stdout="- config:\n  name: default\n",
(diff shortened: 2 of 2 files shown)


craft_providers/lxd/lxc.py
```diff
@@ -793,10 +793,17 @@ def project_create(self, *, project: str, remote: str = "local") -> None:
         try:
             self._run_lxc(command, capture_output=True, project=project)
         except subprocess.CalledProcessError as error:
-            raise LXDError(
-                brief=f"Failed to create project {project!r}.",
-                details=errors.details_from_called_process_error(error),
-            ) from error
+            if project in self.project_list(remote=remote):
```
mr-cal, line 796: Can you add a comment along the lines of this?
`handle the race condition where two processes check and create a project at the same time`

mr-cal (approved):
Looks great!

craft_providers/lxd/lxc.py
syu-w: Sure

craft_providers/lxd/lxc.py
```diff
@@ -793,10 +793,19 @@ def project_create(self, *, project: str, remote: str = "local") -> None:
         try:
             self._run_lxc(command, capture_output=True, project=project)
         except subprocess.CalledProcessError as error:
-            raise LXDError(
-                brief=f"Failed to create project {project!r}.",
-                details=errors.details_from_called_process_error(error),
-            ) from error
+            # handle the race condition where two processes check and
+            # create the same project at the same time
+            if project in self.project_list(remote=remote):
```
beliaev-maksim, line 798: maybe instead of doing a check in exception, we can do a check `if project in self.project_list` before we try to run ` self._run_lxc` or this will not work?

syu-w: This will still have race chance. 2 processes see the project not exists and try create it together. Since the only lock is in the LXC.

mr-cal: Sheng is right - the lock is in LXC.  Craft-providers already checks if the project exists before calling this function, but it occurs in a different module ([here](https://github.com/canonical/craft-providers/blob/bb4ff5a849cb81199b5630099080dbbb0499cecf/craft_providers/lxd/launcher.py#L125-L130))

beliaev-maksim: I am not 100% sure. How deterministic is this approach? 

can we end up that the error will be again false positive due to race?
will LXD return an error if "the project is in creation process" and we still call "create project" ?

syu-w: Create project should be close to atomic and very fast.

mr-cal: Ah yes, that's a fair point.  I'm not sure if LXC will block or fail on `lxc project list` if `lxc project create` is in process.

This approach worked for `lxc remote list` and `lxc remote add`.  

If this race condition continues to occur after this PR or we get a report of another command failing (rule of 3), then I think we will look into other solutions, such as creating a utility function with similar logic plus some retry/backoff timing.

beliaev-maksim: agreed

PR #376, opened 2023-07-31
lxc: create status data in lxd to avoid race

### LINKED ISSUE #341: lxd: race condition when running multiple instances of craft-providers
### Overview

New race condition reported from charmcraft: https://github.com/canonical/charmcraft/issues/1132

### Details

This is similar to an issue from 2022: 
- https://github.com/canonical/charmcraft/issues/554
- #96 
- #158

The charm's CI pipeline runs multiple instances of charmcraft in parallel and create a race condition when executing lxc commands.  In the prior bug, the race condition was caused by `lxc remote add`.  In this new report, the race condition is caused by `lxc project create`.

### Logs

https://github.com/canonical/kfp-operators/actions/runs/5399918594/jobs/9807754324

```
Traceback (most recent call last):
  File "/home/runner/work/kfp-operators/kfp-operators/tests/integration/test_bundle.py", line 39, in test_build_and_deploy
    built_charms = await ops_test.build_charms(*charms_to_build.values())
  File "/home/runner/work/kfp-operators/kfp-operators/.tox/bundle-integration/lib/python3.8/site-packages/pytest_operator/plugin.py", line 1028, in build_charms
    charms = await asyncio.gather(
  File "/home/runner/work/kfp-operators/kfp-operators/.tox/bundle-integration/lib/python3.8/site-packages/pytest_operator/plugin.py", line 1011, in build_charm
    raise RuntimeError(
RuntimeError: Failed to build charm /home/runner/work/kfp-operators/kfp-operators/charms/kfp-viz:
Packing the charm.
Launching environment to pack for base name='ubuntu' channel='20.04' architectures=['amd64'] (may take a while the first time but it's reusable)
charmcraft internal error: LXDError(brief="Failed to create project 'charmcraft'.", details='* Command that failed: \'lxc --project charmcraft project create local:charmcraft\'\n* Command exit code: 1\n* Command standard error output: b\'Error: Failed creating project "charmcraft": Failed adding database record: This "projects" entry already exists\\n\'', resolution=None)
Full execution log: '/home/runner/.local/state/charmcraft/log/charmcraft-20230628-102516.095719.log'
```

### Solution

Craft-providers can use the same race condition handling that was re-used for the prior issue (PR linked above).

syu-w:
Use `user.craft_providers.status` and `user.craft_providers.timer` in LXD to detect current status to avoid race
that try to create same instance from different process.

Also added restart() to instance to avoid ephemeral delete it when stop for configuration.

Fixes #341

### DIFF
--- a/craft_providers/lxd/launcher.py
+++ b/craft_providers/lxd/launcher.py
@@ -22,7 +22,9 @@ import os
 import re
 import subprocess
 import sys
-from datetime import datetime, timedelta
+import threading
+import time
+from datetime import datetime, timedelta, timezone
 from typing import Optional
 
 from craft_providers import Base, ProviderError, bases
@@ -31,9 +33,52 @@ from craft_providers.errors import details_from_called_process_error
 from .errors import LXDError
 from .lxc import LXC
 from .lxd_instance import LXDInstance
+from .lxd_instance_status import ProviderInstanceStatus
 from .project import create_with_default_profile
 
 logger = logging.getLogger(__name__)
+
+
+class InstanceTimer(threading.Thread):
+    """Timer for update instance that still alive."""
+
+    __active: bool = True
+    __interval: float
+    __instance: LXDInstance
+
+    def __init__(self, instance: LXDInstance, interval: int = 3) -> None:
+        """Initialize the timer.
+
+        :param instance: LXD instance to update.
+        :param interval: Interval in seconds to update the instance timer.
+        """
+        self.__instance = instance
+        self.__interval = interval
+        super().__init__(daemon=True)
+
+    def run(self) -> None:
+        """Run the timer."""
+        while self.__active:
+            now = datetime.now(timezone.utc).isoformat()
+            try:
+                self.__instance.config_set("user.craft_providers.timer", now)
+                logger.debug("Set instance timer to %r", now)
+            except LXDError:
+                # Error in timer update is not critical
+                logging.exception("Error updating instance timer")
+            time.sleep(self.__interval)
+
+        try:
+            self.__instance.config_set("user.craft_providers.timer", "DONE")
+            logger.debug("Set instance timer to 'DONE'")
+        except LXDError:
+            # Error in timer update is not critical
+            logging.exception("Error updating instance timer")
+        logger.debug("Instance timer update stopped.")
+
+    def stop(self) -> None:
+        """Stop the timer."""
+        self.__active = False
 
 
 def _create_instance(
@@ -71,49 +116,107 @@ def _create_instance(
     :param lxc: LXC client.
     """
     logger.info("Creating new instance from remote")
-    logger.debug(
-        "Creating new instance from image %r from remote %r.", image_name, image_remote
-    )
-    instance.launch(image=image_name, image_remote=image_remote, ephemeral=ephemeral)
-    base_configuration.setup(executor=instance)
 
-    _set_timezone(instance, project, remote, lxc)
-
-    # return early if base instances and user id mapping are not specified
-    if not base_instance and not map_user_uid:
-        return
-
-    # the instance needs to be stopped before copying or updating the id map
-    instance.stop()
-
+    # Lockable base instance creation. Only one caller can create the base instance.
+    # Other callers will will get LXDError and wait until the base instance is created.
     if base_instance:
-        logger.info("Creating new base instance from instance")
+        logger.info("Creating new base instance from remote")
         logger.debug(
-            "Creating new base instance %r from instance.", base_instance.instance_name
+            "Creating new base instance from image %r from remote %r",
+            image_name,
+            image_remote,
+        )
+        base_instance.launch(
+            image=image_name,
+            image_remote=image_remote,
+            ephemeral=False,  # base instance should not ephemeral
+        )
+        base_instance_status = base_instance.config_get("user.craft_providers.status")
+
+        # Skip the base configuration if the instance is already configured.
+        if base_instance_status != ProviderInstanceStatus.FINISHED.value:
+            logger.debug("Setting up base instance %r", base_instance.instance_name)
+            base_instance.config_set(
+                "user.craft_providers.status", ProviderInstanceStatus.PREPARING.value
+            )
+            config_timer = InstanceTimer(base_instance)
+            config_timer.start()
+            base_configuration.setup(executor=base_instance)
+            _set_timezone(
+                base_instance,
+                base_instance.project,
+                base_instance.remote,
+                base_instance.lxc,
+            )
+            base_instance.config_set(
+                "user.craft_providers.status", ProviderInstanceStatus.FINISHED.value
+            )
+            # set the full instance name as image description
+            lxc.config_set(
+                instance_name=base_instance.instance_name,
+                key="image.description",
+                value=base_instance.name,
+                project=project,
+                remote=remote,
+            )
+            config_timer.stop()
+            base_instance.stop()
+
+        # Copy the base instance to the instance.
+        logger.info("Creating new instance from base instance")
+        logger.debug(
+            "Creating new instance %r from base instance %r",
+            instance.instance_name,
+            base_instance.instance_name,
         )
         lxc.copy(
-            source_remote=remote,
-            source_instance_name=instance.instance_name,
+            source_remote=base_instance.remote,
+            source_instance_name=base_instance.instance_name,
             destination_remote=remote,
-            destination_instance_name=base_instance.instance_name,
+            destination_instance_name=instance.instance_name,
             project=project,
         )
+        _set_timezone(instance, project, remote, lxc)
+    else:
+        logger.debug(
+            "Creating new instance from image %r from remote %r",
+            image_name,
+            image_remote,
+        )
+        instance.launch(
+            image=image_name, image_remote=image_remote, ephemeral=ephemeral
+        )
+        instance_status = instance.config_get("user.craft_providers.status")
 
-        # set the full instance name as image description
-        lxc.config_set(
-            instance_name=base_instance.instance_name,
-            key="image.description",
-            value=base_instance.name,
-            project=project,
-            remote=remote,
-        )
+        # Skip the base configuration if the instance is already configured.
+        if instance_status != ProviderInstanceStatus.FINISHED.value:
+            logger.info("Setting up instance")
+            logger.debug("Setting up instance %r", instance.instance_name)
+            instance.config_set(
+                "user.craft_providers.status", ProviderInstanceStatus.PREPARING.value
+            )
+            config_timer = InstanceTimer(instance)
+            config_timer.start()
+            base_configuration.setup(executor=instance)
+            _set_timezone(instance, project, remote, lxc)
+            instance.config_set(
+                "user.craft_providers.status", ProviderInstanceStatus.FINISHED.value
+            )
+            config_timer.stop()
+            if not ephemeral:
+                # stop ephemeral instances will delete them immediately
+                instance.stop()
 
     # after creating the base instance, the id map can be set
     if map_user_uid:
         _set_id_map(instance=instance, lxc=lxc, project=project, remote=remote, uid=uid)
 
     # now restart and wait for the instance to be ready
-    instance.start()
+    if ephemeral:
+        # ephemeral instances can only be restarted
+        instance.restart()
+    else:
+        instance.start()
     base_configuration.wait_until_ready(executor=instance)
 
 
--- a/craft_providers/lxd/lxc.py
+++ b/craft_providers/lxd/lxc.py
@@ -16,16 +16,22 @@
 #
 
 """LXC wrapper."""
+import contextlib
 import enum
 import logging
 import pathlib
 import shlex
 import subprocess
+import threading
+import time
+from collections import deque
+from datetime import datetime, timezone
 from typing import Any, Callable, Dict, List, Optional
 
 import yaml
 
 from craft_providers import errors
+from craft_providers.lxd.lxd_instance_status import ProviderInstanceStatus
 
 from .errors import LXDError
 
@@ -58,6 +64,7 @@ class LXC:
         lxc_path: pathlib.Path = pathlib.Path("lxc"),
     ) -> None:
         self.lxc_path = lxc_path
+        self.lxc_lock = threading.Lock()
 
     def _run_lxc(
         self,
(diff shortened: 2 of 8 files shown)


craft_providers/lxd/lxc.py
```diff
@@ -995,3 +1032,85 @@ def stop(
...
+        project: str = "default",
+        remote: str = "local",
+    ) -> None:
+        """Check build status of instance."""
+        build_status: Optional[str] = None
+        instance_info: Dict[str, Any] = {"status": ""}
+        retry_count: int = 0
+
+        # Approx max wait time is 2 hours
+        while retry_count < 2400:
```
mr-cal, line 1049: I would prefer a more standard time-based timeout and loop for two reasons:
1. readability - 2,400 retries is hard to comprehend. It took me a while to wrap my head around how long a `retry_count` is (like on the line `if retry_count == 100:`), since I have to divide 2,400 retries by approximately 2 hours
2. predictability - Timeouts are a bit more predictable.  I can imagine the actual timeout of 2,400 retries varying from system to system.

mr-cal (commented):
Looking good so far, thanks for pulling me in early.  I like the strategy.

craft_providers/lxd/lxc.py
```diff
@@ -995,3 +1032,85 @@ def stop(
...
+                and config_status == "DONE"
+                and instance_info["status"] == "STOPPED"
+            ):
+                # Instance is ready
+                return
+
+            # Check if instance is stuck in starting state
+            if (
+                build_status in ("STARTING", "", None)
+                and instance_info["status"] == "STOPPED"
```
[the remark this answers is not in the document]
syu-w: `instance_info["status"]` is from LXD.

craft_providers/lxd/lxc.py
```diff
@@ -995,3 +1032,85 @@ def stop(
...
+            ):
+                if retry_count > 200:
+                    # The instance is stuck in starting state, delete it
+                    with contextlib.suppress(LXDError):
+                        # Ignore errors that someone else already deleted the instance
+                        self.delete(
```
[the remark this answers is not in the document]
syu-w: Sure, added.

craft_providers/lxd/lxc.py
syu-w: I thinking that too, but I move to time + count based.

craft_providers/lxd/lxc.py
```diff
@@ -995,3 +1032,85 @@ def stop(
...
+            except LXDError:
+                if retry_count == 100:
+                    # Something went wrong, unable to get instance info after 5 minutes
+                    raise
+                sleep(3)
+                retry_count += 1
+                continue
+
+            if (
+                build_status == "FINISHED"
```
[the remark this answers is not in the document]
syu-w: Added

craft_providers/lxd/launcher.py
```diff
@@ -36,6 +38,32 @@
...
+class InstanceTimer(threading.Thread):
+    """Timer for update instance that still alive."""
+
+    count: int = 0
+    active: bool = True
+    instance: LXDInstance
+
+    def __init__(self, instance: LXDInstance) -> None:
+        self.instance = instance
+        threading.Thread.__init__(self, daemon=True)
```
lengau, line 50:
```suggestion
        super().__init__(daemon=True)
```

craft_providers/lxd/launcher.py
```diff
@@ -36,6 +38,32 @@
...
+
+    def run(self) -> None:
+        """Run the timer."""
+        while self.active:
+            self.count += 1
+            self.instance.config_set("user.craft_providers.timer", str(self.count))
+            logger.debug("Set instance timer to %r", self.count)
+            time.sleep(3)
```
lengau, line 58: Perhaps we should have the tick period be configurable in `__init__`?

tests/unit/lxd/test_launcher.py
```diff
@@ -143,10 +158,18 @@ def test_launch_no_base_instance(
     assert fake_instance.mock_calls == [
         call.exists(),
         call.launch(image="image-name", image_remote="image-remote", ephemeral=False),
+        call.config_get("user.craft_providers.status"),
+        call.config_get().__ne__("FINISHED"),
```
lengau, line 162: This might be too brittle. It would probably be better to set `fake_instance.return_value` in the arrange space.

You can also test different parts of the loop by setting the fake instance's `side_effect` to a list of return values for successive calls.

lengau (changes requested):

craft_providers/lxd/launcher.py
syu-w: Added

tests/unit/lxd/test_launcher.py
syu-w: I am not sure why, but that is pytest output. And that are call checks not return values.

lengau: `fake_instance` is a `Mock` object, so it's returning child `Mock` objects. If you set `fake_instance.config_get.return_value = "FINISHED"` before running the test, it shouldn't need this line.

Likewise, to check that the instance works correctly when the first query is in-progress but the second query is finished, you can set `fake_instance.config_get.side_effect = ["PREPARING", "FINISHED]`.

syu-w: Done

craft_providers/lxd/launcher.py
```diff
@@ -36,6 +38,37 @@
...
+
+    def __init__(self, instance: LXDInstance, interval: int = 3) -> None:
+        self.instance = instance
+        self.interval = interval
+        super().__init__(daemon=True)
+
+    def run(self) -> None:
+        """Run the timer."""
+        while self.active:
+            self.count += 1
```
lengau, line 56: Question for both you and @mr-cal about this: should we use some sort of date/time value (e.g. an ISO 8601 timestamp) instead for updating the timer? That way it becomes obvious when the timer was last updated.

craft_providers/lxd/launcher.py
```diff
@@ -36,6 +38,37 @@
...
+            try:
+                self.instance.config_set("user.craft_providers.timer", str(self.count))
+                logger.debug("Set instance timer to %r", self.count)
+            except LXDError:
+                # Error in timer update is not critical
+                logging.exception("Error updating instance timer")
+            time.sleep(self.interval)
+
+        logger.debug("Instance timer update stopped.")
```
lengau, line 65: We should probably set the config value here to "done" or another sentinel value.

tests/integration/lxd/test_launcher.py
```diff
@@ -439,6 +440,63 @@ def test_launch_with_project_and_use_base_instance(
...
+                image_name="22.04",
+                image_remote="ubuntu",
+                use_base_instance=True,
+                project=project,
+                remote="local",
+            )
+
+    # launch 5 instances from an image and create a base instance
+    for i in range(0, 5):
+        instance_thread = InstanceThread(f"{instance_name}-{i}")
```
lengau, line 470: BTW this is a very clever way to do this test. I like it!

craft_providers/lxd/launcher.py
syu-w: We can use time if needed, it just need to be changed every x seconds. So put a ISO 8601 should be fine.

craft_providers/lxd/launcher.py
syu-w: We can set it to DONE or something, but it is not used later anyway.

lengau: I'm more thinking for user debugging if something goes wrong — if the timer is marked as done, we know it completed and something else failed. If not, we can presume the process crashed or something.

craft_providers/lxd/launcher.py
mr-cal: I like the idea of a timestamp too, I could see it being useful if we need to debug. 

The count is fine as well.

craft_providers/lxd/launcher.py
```diff
@@ -72,48 +105,96 @@ def _create_instance(
     """
     logger.info("Creating new instance from remote")
     logger.debug(
-        "Creating new instance from image %r from remote %r.", image_name, image_remote
+        "Creating new instance from image %r from remote %r", image_name, image_remote
     )
```
mr-cal, line 109: Should these be moved to line 167, where `instance.launch()` is called?

As it is now, the timing/order of these logs doesn't make sense.

craft_providers/lxd/launcher.py
```diff
@@ -72,48 +105,96 @@ def _create_instance(
...
+        )
+        base_instance.launch(
+            image=image_name,
+            image_remote=image_remote,
+            ephemeral=False,  # base instance should not ephemeral
+        )
+        base_instance_status = base_instance.config_get("user.craft_providers.status")
+
+        # Skip the base configuration if the instance is already configured.
+        if base_instance_status != "FINISHED":
```
mr-cal, line 128: Can you define these statuses in one place and refer to the variable?  (enums are always a good choice for pre-defined states)

You could also make a small class and wrap `base_instance.config_get("user.craft_providers.status")` in a function.

craft_providers/lxd/launcher.py
```diff
@@ -72,48 +105,96 @@ def _create_instance(
...
-        return
-
-    # the instance needs to be stopped before copying or updating the id map
-    instance.stop()
 
+    # Lockable base instance creation. Only one caller can create the base instance.
+    # Other callers will will get LXDError and wait until the base instance is created.
     if base_instance:
-        logger.info("Creating new base instance from instance")
+        logger.info("Preparing base instance from remote")
```
mr-cal, line 114: pedantic: `Creating new` is used in the other log messages

```suggestion
        logger.info("Creating new base instance from remote")
```

craft_providers/lxd/launcher.py
```diff
@@ -72,48 +105,96 @@ def _create_instance(
...
-            remote=remote,
+        _set_timezone(instance, project, remote, lxc)
+    else:
+        instance.launch(
+            image=image_name, image_remote=image_remote, ephemeral=ephemeral
         )
+        instance_status = instance.config_get("user.craft_providers.status")
+
+        # Skip the base configuration if the instance is already configured.
+        if instance_status != "FINISHED":
```
mr-cal, line 174: What scenario could occur for the status to be `FINISHED` here?

craft_providers/lxd/launcher.py
```diff
@@ -72,48 +105,96 @@ def _create_instance(
...
+                # stop ephemeral instances will delete them immediately
+                instance.stop()
 
     # after creating the base instance, the id map can be set
     if map_user_uid:
         _set_id_map(instance=instance, lxc=lxc, project=project, remote=remote, uid=uid)
 
     # now restart and wait for the instance to be ready
-    instance.start()
+    if ephemeral:
```
mr-cal, line 193: Nice, I think you may have fixed a bug with stopping and starting ephemeral instances.

craft_providers/lxd/lxc.py
```diff
@@ -573,18 +590,60 @@ def launch(
...
 
-        try:
-            self._run_lxc(
-                command,
-                capture_output=True,
-                stdin=StdinType.INTERACTIVE,
-                project=project,
+        while retry_count < 3:
+            try:
+                # Try to launch instance
```
mr-cal, line 595: Can you make this comment explain why we need to retry / why it may fail?

mr-cal (commented):
Nice work!

I haven't finished reviewing yet (I need to wrap up for the day) but here are my comments so far.  I'll finish tomorrow.

craft_providers/lxd/launcher.py
syu-w: It could have configuration error or network issues, which the base instance image can be used but stopped at preparing. just need to redo the configuration part.

tests/unit/lxd/test_launcher.py
```diff
@@ -212,19 +236,46 @@ def test_launch_use_base_instance(
...
+    mock_lxd_instance,
+    mock_platform,
+    mock_timezone,
+):
+    """Launch a base instance from an image, but lxc commands fail."""
+    mock_lxc.config_set.side_effect = [
+        LXDError("test1"),
+        LXDError("test2"),
+        LXDError("test3"),
+    ]
```
mr-cal, line 265: This test only calls `config_set` once before the LXDError is raised so why specify 2 more side_effects that are unused?

tests/unit/lxd/test_lxc.py
```diff
@@ -930,6 +950,120 @@ def test_launch(fake_process):
...
+
+    assert mock_check.mock_calls == [
+        call(
+            instance_name="test-instance", project="test-project", remote="test-remote"
+        )
+    ]
+
+
+def test_launch_failed_retry_failed(fake_process, mocker):
+    """Test that we retry launching an instance if it fails, but failed more than 3 times."""
```
mr-cal, line 995: Nice docstrings on these tests

tests/unit/lxd/test_lxc.py
```diff
@@ -984,9 +1159,373 @@ def test_launch_error(fake_process):
...
+
+
+def test_check_instance_status_boot_failed(fake_process, mocker):
+    time_time = mocker.patch("time.time")
+    time_time.return_value = 0
+    mocker.patch("time.sleep")
+    mock_instance = mocker.patch("craft_providers.lxd.lxc.LXC.info")
+    mock_instance.return_value = {"Status": "STOPPED"}
+    mock_instance_config = mocker.patch("craft_providers.lxd.lxc.LXC.config_get")
+    mock_instance_config.side_effect = [
```
mr-cal, line 1251:
```suggestion
    mock_instance_config.side_effect = ["STARTING", 10] * 20
```

sergiusens:
@syu-w any updates on this?

syu-w:
> @syu-w any updates on this?

WIP. Should be ready for review tomorrow.

tests/integration/lxd/test_launcher.py
```diff
@@ -267,7 +268,7 @@ def test_launch_use_base_instance(
         .rstrip("\n")
     )
 
-    assert instance_hostname == base_configuration._hostname
+    assert instance_hostname == instance.instance_name
```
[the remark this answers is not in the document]
syu-w: Because the creation order changed.
Old:
`base_configuration` is the target instance hostname, and we rename it to the base instance hostname later.
New:
`base_configuration` is the base instance hostname, and we rename it to the target instance hostname later.

The final results did not change.

tests/unit/lxd/test_launcher.py
syu-w: It's a leftover from my intermediate tests

mr-cal (approved):
:+1:

craft_providers/lxd/lxc.py
```diff
@@ -573,18 +592,65 @@ def launch(
...
-                command,
-                capture_output=True,
-                stdin=StdinType.INTERACTIVE,
-                project=project,
+        # The total times of retrying to launch the same instance per craft-providers.
+        # If parallel lxc failed, the bad instance will be deleted by the lock holder
+        # or any other craft-providers, and the lock will be released.
+        # However, the new instance lock could be held by any craft-providers.
+        # This is used to avoid lock holder dead and all others are blocked.
+        while retry_count < 3:
```
cmatsuoka, line 600: Are 3 retries enough for typical cases, considering slow machines?

syu-w: It should not have any retry there if lxc is working. But retry when the instance creation failed. So I think 3 times is enough for network error or bad system configuration.

lengau: I'm not too worried about having 3 retries here. The types of errors we want to catch should be ephemeral.

lengau (approved):
Looks good, thanks!

PR #378, opened 2023-08-11
lxd: update logging semantics

### LINKED ISSUE #377: Add INFO level log messages
### What needs to get done

We should either add new user-level INFO messages or change the current WARNING ones to INFO.

### Why it needs to get done

As part of the process to unify the message presenting among the craft-libs.

tigarmo:
This commit does two things:

- Update almost all instances of log.warning() to log.debug() in lxd- related code. The rationale is that the majority of those messages are meant to assist developers in debugging/troubleshooting, and not really meant for end-users.
- Add a couple of log.info() messages. These are only a couple on purpose, as the intent is to provide a _bit_ more information during instance fetching/setup without overwhelming the end-user with info that is ultimately not relevant to them ("implementation details"). For example, we say that we're "Starting the instance" but not that we are "Configuring networking". Since the debug messages are preserved, the information is ultimately still available in logfiles.

Note that multipass code still needs to be updated.

Fixes #377

-----

### DIFF
--- a/craft_providers/lxd/launcher.py
+++ b/craft_providers/lxd/launcher.py
@@ -70,7 +70,8 @@ def _create_instance(
     :param remote: LXD remote to create instance on.
     :param lxc: LXC client.
     """
-    logger.warning(
+    logger.info("Creating new instance from remote")
+    logger.debug(
         "Creating new instance from image %r from remote %r.", image_name, image_remote
     )
     instance.launch(image=image_name, image_remote=image_remote, ephemeral=ephemeral)
@@ -86,7 +87,8 @@ def _create_instance(
     instance.stop()
 
     if base_instance:
-        logger.warning(
+        logger.info("Creating new base instance from instance")
+        logger.debug(
             "Creating new base instance %r from instance.", base_instance.instance_name
         )
         lxc.copy(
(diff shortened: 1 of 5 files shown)


mr-cal (approved):

mr-cal:
Whoops, forgot it was still a draft.  Have an early approval, free of charge.

craft_providers/lxd/launcher.py
```diff
@@ -630,13 +630,14 @@ def launch(
...
-        "Creating instance from base instance %r", base_instance.instance_name
+    logger.info("Creating instance from base instance")
+    logger.debug(
+        "Creating instance from base instance %r.", base_instance.instance_name
     )
 
     # the base instance is not expected to be running but check for safety
     if base_instance.is_running():
-        logger.warning("Stopping base instance.")
+        logger.debug("Stopping base instance.")
```
tigarmo, line 640: @mr-cal while trying to cover this line I realized that we log that we're stopping the base instance but we don't actually call `base_instance.stop()` here. Is that right?

mr-cal: Good catch.  I think you're right - the base instance should be stopped here.

On a related note, https://github.com/canonical/craft-providers/pull/376 is supposed going to flip the order of creation (previously: creating an instance and copying to the base instance new: create a base instance and copy it to the instance).

I'll keep this in mind when I review that PR.  If this is still outstanding, I'll make an issue (or just fix it).

tigarmo: here's what I propose: I don't think we should make that kind of change in this "logging" PR, so I'll add the test checking that `base_instance.stop()` is _not_ called. Then when we get around to fixing it we have an existing, breaking test that just needs to be updated. Is that ok?

mr-cal: Yes!

syu-w: I think you can just leave it there for now, I will rebase my PR on top of that. So you don't need to keep tracking what I am changing.

tigarmo: thanks!

Issue #184, opened 2022-12-07
Use `snap refresh --hold`


mr-cal:
This https://github.com/canonical/craft-providers/pull/182 adds support for pausing automatic snap refreshes.

It has to run `snap set system refresh.hold=<24h>` during every `setup()` and `warmup()` call.

It would be more efficient to run `snap refresh --hold` once during `setup()`.  This feature of snapd is not yet stable.  Once it becomes stable, craft-providers should use this command instead.

mr-cal:
https://github.com/canonical/charmcraft/issues/1202 proves that the snap is being refreshed despite the `snap set system refresh.hold=<24h>` call.  The workaround was to use the new syntax `snap refresh --hold`.

If more issues continue to appear, perhaps craft-providers should explicitly check what revision of a snap is installed, rather than trusting the revision recorded in the craft-instance.conf file.

sergiusens:
we want to update the compatibility tag

PR #380, opened 2023-08-15
tests: refactor verify_network and DEFAULT_FAKE_CMD usage

mr-cal:

-----

Some minor test refactoring in preparation for #184.

(CRAFT-1984)

### DIFF
--- a/tests/unit/bases/test_almalinux.py
+++ b/tests/unit/bases/test_almalinux.py
@@ -35,7 +35,7 @@ from craft_providers.instance_config import InstanceConfiguration
 from logassert import Exact  # type: ignore
 from pydantic import ValidationError
 
-DEFAULT_FAKE_CMD = ["fake-executor"]
+from tests.unit.conftest import DEFAULT_FAKE_CMD
 
 
 @pytest.fixture()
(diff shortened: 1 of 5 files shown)


syu-w (approved):

tests/unit/conftest.py
```diff
@@ -179,3 +176,12 @@ def home_temporary_file():
         home_temporary_file,
     )
     return temp_file
+
+
+@pytest.fixture()
+def mock_verify_network(fake_process):
```
lengau, line 182: To get **very** technical, this is neither a fake nor a mock, but a stub. (It returns a real, but predefined, value.)

(But also, it's not like this distinction is actually particularly useful here.)

tests/unit/bases/test_almalinux.py
```diff
@@ -811,19 +811,14 @@ def test_setup_snapd_proxy_failures(fake_process, fake_executor, fail_index):
 
 
 @pytest.mark.parametrize("fail_index", list(range(0, 2)))
-def test_pre_setup_snapd_failures(fake_process, fake_executor, fail_index):
+def test_pre_setup_snapd_failures(
+    mock_verify_network, fake_process, fake_executor, fail_index
+):
```
lengau, line 816: As a preventative against linters later complaining about `mock_verify_network` being unused, you could do:

```suggestion
@pytest.mark.usefixtures("mock_verify_network")
def test_pre_setup_snapd_failures(fake_process, fake_executor, fail_index):
```

We currently have the annotation checks disabled in ruff for tests, which is [where ruff organises this check](https://beta.ruff.rs/docs/rules/unused-function-argument/). I don't really see a need to enable it for tests, but if we were to do so later this would flag.

lengau (approved):
Nice cleanup, thanks!

I added what might be exaggerating to call nitpicks.

tests/unit/bases/test_almalinux.py
mr-cal: I like the preventative work, thanks!

PR #382, opened 2023-08-16
bases: use `snap refresh --hold`

### LINKED ISSUE #184: Use `snap refresh --hold`

mr-cal:

-----

### Overview
- Replace `snap set system refresh.hold=<24h>` with `snap refresh --hold`
- Execute this call only during `setup` instead of `setup` and `warmup`
- Bump compatibility tag

### Details

This required moving a few other snapd-related calls to only execute during `setup`. 

In `craft-providers<1.13.0`, these snapd calls were [only made once](https://github.com/canonical/craft-providers/blob/3ea09344578f03a458676fac51784bbe6ea7f305/craft_providers/bases/ubuntu.py#L378-L379) during `setup`.  During the [base refactor](https://github.com/canonical/craft-providers/pull/274) in `1.13.0`, these calls were made during `setup` and `warmup`.  I've reverted to the previous behavior because it works, speeds up warmup, and is a neat and tidy way to make this change.

Compatibility tag bump is in a separate commit for your viewing pleasure.

Resolves #184 

https://github.com/canonical/charmcraft/issues/1202
(CRAFT-1984)

### DIFF
--- a/tests/unit/bases/test_almalinux.py
+++ b/tests/unit/bases/test_almalinux.py
@@ -41,7 +41,7 @@ from tests.unit.conftest import DEFAULT_FAKE_CMD
 def mock_load(mocker):
     return mocker.patch(
         "craft_providers.instance_config.InstanceConfiguration.load",
-        return_value=InstanceConfiguration(compatibility_tag="almalinux-base-v1"),
+        return_value=InstanceConfiguration(compatibility_tag="almalinux-base-v2"),
     )
 
 
@@ -150,7 +150,7 @@ def mock_get_os_release(mocker):
     ],
 )
 @pytest.mark.parametrize(
-    ("tag", "expected_tag"), [(None, "almalinux-base-v1"), ("test-tag", "test-tag")]
+    ("tag", "expected_tag"), [(None, "almalinux-base-v2"), ("test-tag", "test-tag")]
 )
 def test_setup(
     fake_process,
@@ -654,7 +654,7 @@ def test_ensure_image_version_compatible_failure(fake_executor, monkeypatch):
         base_config._ensure_instance_config_compatible(executor=fake_executor)
 
     assert exc_info.value == BaseCompatibilityError(
-        "Expected image compatibility tag 'almalinux-base-v1', found 'invalid-tag'"
+        "Expected image compatibility tag 'almalinux-base-v2', found 'invalid-tag'"
     )
 
 
@@ -1057,7 +1057,7 @@ def test_update_setup_status(fake_executor, mock_load, status):
     assert fake_executor.records_of_push_file_io == [
         {
             "content": (
-                "compatibility_tag: almalinux-base-v1\n"
+                "compatibility_tag: almalinux-base-v2\n"
                 f"setup: {str(status).lower()}\n".encode()
             ),
             "destination": "/etc/craft-instance.conf",
@@ -1163,7 +1163,7 @@ def test_ensuresetup_completed_not_setup(status, fake_executor, mock_load):
 )
 def test_warmup_overall(environment, fake_process, fake_executor, mock_load, mocker):
     mock_load.return_value = InstanceConfiguration(
-        compatibility_tag="almalinux-base-v1", setup=True
+        compatibility_tag="almalinux-base-v2", setup=True
     )
     alias = almalinux.AlmaLinuxBaseAlias.NINE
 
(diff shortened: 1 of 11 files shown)


tests/unit/bases/test_almalinux.py
```diff
@@ -843,72 +829,60 @@ def test_pre_setup_snapd_failures(fake_process, fake_executor, fail_index):
 
 
 @pytest.mark.usefixtures("stub_verify_network")
-def test_setup_snapd_failures(fake_process, fake_executor):
+@pytest.mark.parametrize("fail_index", list(range(0, 7)))
+def test_setup_snapd_failures(fake_process, fake_executor, fail_index, mocker):
```
mr-cal, line 833: Diff is pretty worthless here. The change here is that 6 snapd calls were moved from `post_setup_snapd` (called during `setup` and `warmup`) to `setup_snapd` (called during `setup`).

craft_providers/bases/almalinux.py
```diff
@@ -211,6 +211,9 @@ def _setup_snapd(self, executor: Executor) -> None:
                 details=details_from_called_process_error(error),
             ) from error
 
+        self._enable_snapd_service(executor=executor)
```
tigarmo, line 214: So if I understand this correctly, changing these calls from `_post_setup_snapd()` to `_setup_snapd` had the consequence of every subclass of `Base` needing to repeat these calls in their implementation of `_setup_snapd`, right? 

I agree with the overall intent of doing the hold only once, but this repetition is unfortunate because it's the kind of thing that's easy to miss when a new base comes along, or if we need to refactor this (will need to remember to update all current implementations). I don't have a "great" idea to address this, the only one that comes to mind is a new method (like `_configure_snapd()`) that gets called between the `_setup_snapd` and `_post_setup_snapd` and doesn't get called in warmup.

Also it was a bit unclear to understand that `_post_setup_snapd()` is called both at setup() and at warmup(). Can we update its docstring to make this more evident?

tigarmo (commented):
thanks! Left two notes, one is just a future thought but the other I really think we should consider.

craft_providers/bases/almalinux.py
syu-w: I think repeat call `snap refresh --hold` shouldn't have issue? So keep them in the `_post_setup_snapd` should be OK?

tests/unit/lxd/test_launcher.py
```diff
@@ -30,7 +30,7 @@
 @pytest.fixture()
 def mock_base_configuration():
     mock_base = Mock(spec=Base)
-    mock_base.compatibility_tag = "mock-compat-tag-v100"
+    mock_base.compatibility_tag = "mock-compat-tag-v200"
```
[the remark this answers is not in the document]
mr-cal: Agreed!  Here's an issue:
https://github.com/canonical/craft-providers/issues/383

craft_providers/bases/almalinux.py
mr-cal: Yeah, I think I shouldn't have tried to do 2 things in 1 PR (use the new call and speed up the warmup time).

I agree that this approach is more error prone and I don't have a good idea either.  Perhaps `warmup` should make a few direct calls (`setup_snapd_proxy()`) instead of calling the high-level setup steps (`post_setup_snapd`).

Let's talk during standup tomorrow.

tests/unit/bases/test_almalinux.py
mr-cal: The diff is now readable

craft_providers/bases/almalinux.py
mr-cal: Thanks for your patience with explaining, I have added what we discussed today.  It's made the PR cleaner and improved warmup time.  

I'm still thinking about a future improvement for `warmup`.  I'm wondering about a series of warmup calls like:
```
def warmup():
  _warmup_image()
  _warmup_snapd()
  _warmup_snaps()
```

tigarmo (approved):

lengau (approved):
Looks good, but please put issues on the appropriate applications re: their configure hooks :-)

syu-w:
I think keep `_post_setup_snapd` will be better for possible future extensions.

mr-cal:
> I think keep `_post_setup_snapd` will be better for possible future extensions.

I'm going to leave this as we agreed during standup (`_post_setup_snapd` only runs during setup, `warmup_snapd` during warmup) and we can iterate later.

syu-w:
> > I think keep `_post_setup_snapd` will be better for possible future extensions.
> 
> I'm going to leave this as we agreed during standup (`_post_setup_snapd` only runs during setup, `warmup_snapd` during warmup) and we can iterate later.

Sounds good

PR #386, opened 2023-08-25
lxc: retry apply user id map if not in effect

syu-w:

-----


craft_providers/lxd/launcher.py
```diff
@@ -486,15 +486,41 @@ def _set_id_map(
...
 
-    lxc.config_set(
-        instance_name=instance.instance_name,
-        key="raw.idmap",
-        value=f"both {uid!s} 0",
-        project=project,
-        remote=remote,
+    # Retry for idmap to be set
+    for _ in range(10):
+        lxc.config_set(
```
mr-cal, line 495: Did you figure out if the problem that setting the config is not working?  Or that it takes time for it to take effect?

Phrased another way - what happens if you execute `config_set` once by moving it outside of the retry loop?  I'm hesitant about calling `config_set` up to 10 times.

syu-w: No, `config_set` return no error, but the value may lost for some reasons. Add sleep does not work.
For what I tested, `config_set` need to be called again if the value not there.

mr-cal: I see.  I think we should tie in the LXD team (perhaps raising a bug with a reproducer with them).

mr-cal:
Oh no!  Is the lock not working?  I see the etag error:
https://github.com/canonical/craft-providers/actions/runs/5979275321/job/16223023940?pr=386#step:7:513

syu-w:
I see that too. Is there possible multiple LXC class used? The lock does not work cross LXC class instance.

craft_providers/lxd/launcher.py
cmatsuoka: @tomponline probably has good advice on how to proceed here.

mr-cal:
That could be as there are a lot of calls that have a default argument of `lxc = LXC()`, both in the code and the tests (try `grep -r "LXC()" -C1`).

craft_providers/lxd/launcher.py
tomponline: In a nutshell are you not getting "read your own writes" behaviour when setting and then getting config entries?

syu-w: I think I fixed it by using only 1 lxc lock in the process. But I there is possible a LXD race that when 2 `lxc config set` running at the same time, one of them return success and the other one return failed, but both failed actually.

tomponline: Are both lxc config processes running against the same instance?

syu-w: Yes, and it usually had a `ETag doesn't match` error. But I don't know how one of them reported success but the value doesn't get updated.

tomponline: Ah right, yes, I think it may be possible, at least on some routes, to not get the etag doesn't match (as the original content was loaded for both requests) and then have them both apply different settings and overwrite one and other. Really I think we should prevent concurrent update requests by using server side locking.

Please can you open an issue.

tomponline: Out of interest, why do you have multiple requests modifying the same instance concurrently?

syu-w: Because we are going to support remote LXD build, and some of the process will use a shared base instance which we want to do a heartbeat on it. Its processing may be interrupted, so we don't want other processes to wait forever even if the LXC status is RUNNING, but it will not be finished unless another one restart the configuration process.

syu-w: https://github.com/canonical/lxd/issues/12189

tomponline: > Because we are going to support remote LXD build, and some of the process will use a shared base instance which we want to do a heartbeat on it. Its processing may be interrupted, so we don't want other processes to wait forever even if the LXC status is RUNNING, but it will not be finished unless another one restart the configuration process.

I'm afraid I didn't understand much of that. `shared base instance`, `heartbeat` etc. :)

syu-w: All *craft (snap / charm) create its own template (the shared base instance). Like it download Ubuntu 22.04 on amd64, install packages and configuration. Then copy the base instance to a new instances to build artifacts for each snap/charm. This has an issue when create multiple artifacts at the same time when the base instance is not exist. Only one process should create the base instance and all others should wait it finished. So we update the status info in the LXC user metadata (`user.*`), which can happened concurrently when we also update the main LXC configuration (like change the UID map).

tomponline: OK thanks for clarifying. We are working on https://github.com/canonical/lxd/pull/12191 to fix this.

craft_providers/lxd/launcher.py
```diff
@@ -486,16 +485,35 @@ def _set_id_map(
...
 
+    # Retry for idmap to be set
     lxc.config_set(
         instance_name=instance.instance_name,
         key="raw.idmap",
         value=f"both {uid!s} 0",
         project=project,
         remote=remote,
     )
+    configured_id_map = lxc.config_get(
```
mr-cal, line 500: I wonder if we should add a comment explaining that we do `config_set()` followed by `config_get()` to catch the bug in https://github.com/canonical/lxd/issues/12189 /  https://github.com/canonical/lxd/pull/12191.

mr-cal (approved):

craft_providers/lxd/launcher.py
```diff
@@ -486,16 +485,35 @@ def _set_id_map(
     :param remote: LXD remote to create instance on.
     :param uid: The uid to be mapped. If not supplied, the current user's uid is used.
     """
+    configured_id_map = ""
     if uid is None:
         uid = os.getuid()
 
+    # Retry for idmap to be set
```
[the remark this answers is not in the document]
syu-w: Yes

craft_providers/lxd/launcher.py
syu-w: I can add some comments on that

tigarmo (approved):

PR #394, opened 2023-09-09
feat(base): cache pip directory

### LINKED ISSUE #397: Add global caching for pip and apt
### What needs to get done

- fix apt issues in https://github.com/canonical/craft-providers/pull/394
- land https://github.com/canonical/craft-providers/pull/394

### Why it needs to get done

n/a

lengau:

-----
Fixes #397

### DIFF
--- a/craft_providers/base.py
+++ b/craft_providers/base.py
@@ -792,12 +792,11 @@ class Base(ABC):
         """
         return
 
-    def _mount_cache_dirs(self, executor: Executor) -> None:
-        """Mount cache directories for this base.
+    def _mount_shared_cache_dirs(self, executor: Executor) -> None:
+        """Mount shared cache directories for this base.
 
         e.g.
-            user cache (normally $HOME/.cache)
-            apt cache
+            pip cache (normally $HOME/.cache/pip)
 
         This will only be run if caching is enabled for this instance.
 
@@ -807,17 +806,36 @@ class Base(ABC):
         if self._cache_path is None:
             logger.debug("No cache path set, not mounting cache directories.")
             return
-        cache_home_proc = executor.execute_run(
+
+        # Get the real path with additional tags.
+        host_base_cache_path = self._cache_path.resolve().joinpath(
+            self.compatibility_tag, str(self.alias)
+        )
+
+        try:
+            host_base_cache_path.mkdir(parents=True, exist_ok=True)
+        except OSError as error:
+            raise BaseConfigurationError(
+                brief=f"Failed to create host cache directory: {host_base_cache_path}"
+            ) from error
+
+        guest_cache_proc = executor.execute_run(
             ["bash", "-c", "echo -n ${XDG_CACHE_HOME:-${HOME}/.cache}"],
             capture_output=True,
-            text=True
+            text=True,
         )
-        user_cache_path = self._cache_path / "user_cache"
-        user_cache_path.mkdir(parents=True, exist_ok=True)
-        executor.mount(
-            host_source=user_cache_path,
-            target=pathlib.PurePosixPath(cache_home_proc.stdout)
+        guest_base_cache_path = pathlib.Path(guest_cache_proc.stdout)
+
+        # PIP cache
+        host_pip_cache_path = host_base_cache_path / "pip"
+        host_pip_cache_path.mkdir(parents=True, exist_ok=True)
+
+        guest_pip_cache_path = guest_base_cache_path / "pip"
+        executor.execute_run(
+            ["mkdir", "-p", guest_pip_cache_path.as_posix()],
         )
+
+        executor.mount(host_source=host_pip_cache_path, target=guest_pip_cache_path)
 
     def _pre_setup_packages(self, executor: Executor) -> None:
         """Do anything before setting up the packages.
@@ -997,6 +1015,8 @@ class Base(ABC):
 
         self._update_compatibility_tag(executor=executor)
 
+        self._mount_shared_cache_dirs(executor=executor)
+
         self._pre_setup_os(executor=executor)
         self._setup_os(executor=executor)
         self._post_setup_os(executor=executor)
@@ -1006,9 +1026,6 @@ class Base(ABC):
         self._pre_setup_network(executor=executor)
         self._setup_network(executor=executor)
         self._post_setup_network(executor=executor)
-
-        if self._cache_path is not None:
-            self._mount_cache_dirs(executor=executor)
 
         self._setup_wait_for_network(executor=executor)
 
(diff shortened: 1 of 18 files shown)


craft_providers/base.py
```diff
@@ -786,6 +792,63 @@ def _post_setup_network(self, executor: Executor) -> None:
         """
         return
 
+    def _mount_shared_cache_dirs(self, executor: Executor) -> None:
+        """Mount a shared cache directories for this base.
```
lengau, line 796:
```suggestion
        """Mount shared cache directories for this base.
```

craft_providers/base.py
```diff
@@ -786,6 +792,63 @@ def _post_setup_network(self, executor: Executor) -> None:
...
+        cache directories are in unusual locations.
+        """
+        if self._cache_path is None:
+            logger.debug("No cache path set, not mounting cache directories.")
+            return
+
+        # Get the real path with additional tags.
+        host_base_cache_path = pathlib.Path(
+            os.path.realpath(self._cache_path), self.compatibility_tag, str(self.alias)
+        )
```
lengau, line 813:
```suggestion
        host_base_cache_path = self._cache_path.resolve().joinpath(
            self.compatibility_tag, str(self.alias)
        )
```

craft_providers/base.py
```diff
@@ -786,6 +792,63 @@ def _post_setup_network(self, executor: Executor) -> None:
...
+        This step should usually be extended, but may be overridden if common
+        cache directories are in unusual locations.
+        """
+        if self._cache_path is None:
+            logger.debug("No cache path set, not mounting cache directories.")
+            return
+
+        # Get the real path with additional tags.
+        host_base_cache_path = pathlib.Path(
+            os.path.realpath(self._cache_path), self.compatibility_tag, str(self.alias)
```
sergiusens, line 812: why os.path.realpath?

craft_providers/base.py
```diff
@@ -786,6 +792,63 @@ def _post_setup_network(self, executor: Executor) -> None:
...
+        )
+
+        try:
+            host_base_cache_path.mkdir(parents=True, exist_ok=True)
+            host_base_cache_path.chmod(0o777)
```
lengau, line 817: I don't think we need this for most caches — lxd and multipass should do the account mapping for us.

craft_providers/base.py
```diff
@@ -786,6 +792,63 @@ def _post_setup_network(self, executor: Executor) -> None:
...
+            ["bash", "-c", "echo -n ${XDG_CACHE_HOME:-${HOME}/.cache}"],
+            capture_output=True,
+            text=True,
+        )
+        guest_base_cache_path = pathlib.Path(guest_cache_proc.stdout)
+
+        # PIP cache
+        host_pip_cache_path = host_base_cache_path / "pip"
+        host_pip_cache_path.mkdir(parents=True, exist_ok=True)
+        host_pip_cache_path.chmod(0o777)
```
lengau, line 843: Likewise here, we shouldn't need to mess with permissions.

tests/integration/multipass/test_multipass_provider.py
```diff
@@ -44,26 +44,36 @@ def test_create_environment(installed_multipass, instance_name):
     assert test_instance.exists() is False
 
 
-@pytest.mark.parametrize("alias", set(BuilddBaseAlias) - {BuilddBaseAlias.XENIAL})
+ALIASES = list(BuilddBaseAlias)
+ALIASES.remove(BuilddBaseAlias.XENIAL)
```
lengau, line 48: In case anyone's curious why I made this change: doing this allows the use of `pytest-xdist` for these tests, since it gives us a guaranteed order.

tests/unit/bases/test_ubuntu_buildd.py
```diff
@@ -686,6 +685,34 @@ def test_ensure_image_version_compatible_failure(fake_executor, monkeypatch):
...
+    )
+
+    base._mount_shared_cache_dirs(fake_executor)
+
+    expected_mounts = [
+        {
+            "host_source": cache_path
+            / base.compatibility_tag
+            / str(base.alias)
+            / "pip",
```
lengau, line 709: This isn't easy to read. Can we make `host_source` its own variable above so it's clearer?

craft_providers/base.py
syu-w: To avoid someone pass a relative path that cannot find when mounting.

craft_providers/base.py
syu-w: Just tested, inside the lxc it will not get mapped.

```
root@craft-buildd-instance:~# cd .cache/
root@craft-buildd-instance:~/.cache# ls -la
total 10
drwxr-xr-x 3 root   root    3 Sep 21 08:15 .
drwx------ 5 root   root    7 Sep 21 08:15 ..
drwxrwxrwx 2 nobody nogroup 2 Sep 21 08:15 pip
root@craft-buildd-instance:~/.cache# 
```

craft_providers/base.py
```diff
@@ -1029,6 +1082,8 @@ def warmup(
         self._image_check(executor=executor)
         self._post_image_check(executor=executor)
 
+        self._mount_shared_cache_dirs(executor=executor)
```
sergiusens, line 1085: @mr-cal do we need to do this here if done in setup?

sergiusens (approved):

craft_providers/base.py
mr-cal: We need to do it every time since I believe the mount is removed when the instance is stopped.  

Sheng, do you know if the mounts are removed when lxd and multipass instances are stopped?

tests/integration/multipass/test_multipass_provider.py
mr-cal: Nice!

craft_providers/__init__.py
```diff
@@ -23,7 +23,7 @@
 from .provider import Provider
 
 try:
-    from ._version import __version__
+    from ._version import __version__  # type: ignore
```
mr-cal, line 26: You should be able to pull this out now that #404 landed

craft_providers/base.py
syu-w: The way we create the mount does not persistent mount/disk in the config. So it has to be remount every time it starts.

mr-cal (approved):

mr-cal:
For the multipass test suite [problem](https://github.com/canonical/craft-providers/actions/runs/6269775514/job/17026702903?pr=394#step:8:359), I think you might be able to solve it with:

```diff
diff --git a/tests/integration/lxd/conftest.py b/tests/integration/lxd/conftest.py
index 7aed2cc..94ecb66 100644
--- a/tests/integration/lxd/conftest.py
+++ b/tests/integration/lxd/conftest.py
@@ -147,7 +147,7 @@ def project(lxc, project_name):


 @pytest.fixture(scope="session")
-def session_project():
+def session_project(installed_lxd):
     lxc = LXC()
     project_name = "craft-providers-test-session"
     lxc_project.create_with_default_profile(lxc=lxc, project=project_name)
```

PR #415, opened 2023-09-26
fix(lxd): fast error on non-creation related errors

syu-w:
Do not wait if the error is not related.

Fix: https://github.com/canonical/craft-providers/issues/413
CRAFT-2067

### DIFF
--- a/craft_providers/lxd/lxc.py
+++ b/craft_providers/lxd/lxc.py
@@ -617,8 +617,12 @@ class LXC:
                     retry_count,
                 )
                 logger.debug(str(error))
-                # Ignore first 3 failed attempts
-                if retry_count >= 2:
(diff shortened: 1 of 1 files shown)


craft_providers/lxd/lxc.py
```diff
@@ -614,7 +614,9 @@ def launch(  # noqa: PLR0912
                 )
                 logger.debug(str(error))
                 # Ignore first 3 failed attempts
-                if retry_count >= 2:
+                if retry_count >= 2 or (
```
mr-cal, line 617: Can you update the `# Ignore first 3 failed attempts` comment also explain why we can ignore "create" in the error message?

mr-cal (approved):

Issue #412, opened 2023-09-25
Parallel building report incompatible base


syu-w:
### Bug Description

LXD reported `Incompatible base detected: instance is marked as not setup.`

When parallel building, the later process sometime report:
```
Incompatible base detected: instance is marked as not setup.
Recommended resolution: Clean incompatible instance and retry the requested operation.
```

### To Reproduce

N/A

### Relevant log output

```shell
Starting charmcraft version 2.4.1.post49+git1f7acce
Logging execution to '/root/.local/state/charmcraft/log/charmcraft-20230925-180426.678371.log'
Packing the charm.
Launching environment to pack for base name='ubuntu' channel='22.04' architectures=['amd64'] (may take a while the first time but it's reusable)
Creating instance from base instance
Starting instance
Incompatible base detected: instance is marked as not setup.
Recommended resolution: Clean incompatible instance and retry the requested operation.
Full execution log: '/root/.local/state/charmcraft/log/charmcraft-20230925-180426.678371.log'
Creating new instance from base instance
```
```

PR #416, opened 2023-09-26
fix(lxd): check another path before use base instance

syu-w:
Fix another code path that may copy unfinshed base instance.

Fix: https://github.com/canonical/craft-providers/issues/412
CRAFT-2076

### DIFF
--- a/craft_providers/lxd/launcher.py
+++ b/craft_providers/lxd/launcher.py
@@ -755,6 +755,13 @@ def launch(
         )
         return instance
 
+    # check if the base instance is still being created
+    base_instance.lxc.check_instance_status(
+        instance_name=base_instance.instance_name,
+        project=project,
+        remote=remote,
+    )
+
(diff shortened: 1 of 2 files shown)


craft_providers/lxd/launcher.py
```diff
@@ -755,6 +755,13 @@ def launch(
         )
         return instance
 
+    # check if the base instance is still creating
```
mr-cal, line 758:
```suggestion
    # check if the base instance is still being creating
```

mr-cal (approved):

craft_providers/lxd/launcher.py
```diff
@@ -755,6 +755,13 @@ def launch(
         )
         return instance
 
+    # check if the base instance is still being creating
```
mr-cal, line 758:
```suggestion
    # check if the base instance is still being created
```

Sorry for suggesting a typo!  We can just leave this as-is so we don't have to wait on CI

PR #444, opened 2023-11-07
feat(snaps): inject base snap from host

syu-w:
Also reuse the base when inject other snaps.

### DIFF
--- a/craft_providers/actions/snap_installer.py
+++ b/craft_providers/actions/snap_installer.py
@@ -340,7 +340,15 @@ def inject_from_host(*, executor: Executor, snap_name: str, classic: bool) -> No
             classic,
         )
 
-    host_revision = get_host_snap_info(snap_name)["revision"]
+    host_snap_info = get_host_snap_info(snap_name)
+    host_snap_base = host_snap_info.get("base", None)
+    if host_snap_base:
+        logger.debug(
+            "Installing base snap %r for %r from host", host_snap_base, snap_name
+        )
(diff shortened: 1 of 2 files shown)


craft_providers/base.py
```diff
@@ -655,6 +655,7 @@ def _install_snaps(self, executor: Executor) -> None:
                     ),
                 )
 
+            # this is only used when environment is set.
```
mr-cal, line 658: I'm not sure if this belongs inside craft-providers. It explains business logic about how a craft app could decide to set this parameter.

```suggestion
```

craft_providers/actions/snap_installer.py
```diff
@@ -340,7 +340,13 @@ def inject_from_host(*, executor: Executor, snap_name: str, classic: bool) -> No
             classic,
         )
 
-    host_revision = get_host_snap_info(snap_name)["revision"]
+    host_snap_info = get_host_snap_info(snap_name)
+    host_snap_base = host_snap_info.get("base", None)
+    if host_snap_base:
+        logger.debug("Installing base snap %r from host", host_snap_base)
```
mr-cal, line 346: This may add useful context to the logs:

```suggestion
        logger.debug("Installing base snap %r for %r from host", host_snap_base, snap_name)
```

mr-cal (changes requested):
Nice work! I only have pedantic feedback.

mr-cal (approved):
:+1: Thanks! Looks like the unit tests need a small fix.

sergiusens (approved):

PR #457, opened 2023-11-21
ci(tests): use simplify intergration tests for macos

syu-w:


tox.ini
```diff
@@ -55,6 +55,16 @@ change_dir =
...
 
+[testenv:integration-simple-py3.{8,9,10,11,12}]  # Configuration for all tests using pytest
+base = testenv, test
+description =
+    integration: Run simple integration tests with pytest
+labels =
+    integration-simple-py3.{8,10,11}: integration-tests-simple
+change_dir =
+    integration: tests/integration-simple
+commands = pytest {tty:--color=yes} --cov --cov-report=xml:{tox_root}/results/coverage-{env_name}.xml --junit-xml={tox_root}/results/test-results-{env_name}.xml {posargs}
```
lengau, line 66: Why a separate directory as opposed to using pytest marks to either include or exclude certain tests? (e.g. a "slow" mark)

mr-cal: +1.  Also, can we call it `smoketest` not `simple`?

PR #458, opened 2023-11-22
tests: use smoke intergration tests for macOS

syu-w:


mr-cal (approved):
Nice work!

tests/integration/multipass/test_basic.py
```diff
@@ -0,0 +1,108 @@
...
+        assert proc.stdout == test_file.read_bytes()
+
+        instance.unmount(target=target)
+        assert instance.is_mounted(host_source=home_tmp_path, target=target) is False
+
+        # test stop instance
+        instance.stop()
+        assert instance.is_running() is False
+    finally:
+        instance.delete()
```
tigarmo, line 105: minor but there's an alternative to a huge try/finally block like this in pytest tests: add the `request` fixture and then do `request.addfinalizer(instance.delete)` after creating the instance

tox.ini
```diff
@@ -53,7 +53,19 @@ labels =
 change_dir =
     unit: tests/unit
     integration: tests/integration
-commands = pytest {tty:--color=yes} --cov --cov-report=xml:{tox_root}/results/coverage-{env_name}.xml --junit-xml={tox_root}/results/test-results-{env_name}.xml {posargs}
+commands =
+    unit: pytest {tty:--color=yes} --cov --cov-report=xml:{tox_root}/results/coverage-{env_name}.xml --junit-xml={tox_root}/results/test-results-{env_name}.xml {posargs}
+    integration: pytest {tty:--color=yes} --cov --cov-report=xml:{tox_root}/results/coverage-{env_name}.xml --junit-xml={tox_root}/results/test-results-{env_name}.xml -m "not smoketest" {posargs}
```
tigarmo, line 58: how come we don't want to run the smoketest in non-macos?

tigarmo (approved):
thanks! Left some minor notes but feel free to ignore

tests/integration/multipass/test_basic.py
syu-w: I think I cannot call `assert instance.exists() is False` if use that?

tox.ini
syu-w: Because that smoketest only for macos? The integration tests in linux usually works fine.

Issue #450, opened 2023-11-08
Timer thread logging is overly verbose


mr-cal:
### What needs to get done

Log less information in the timer thread.

### Why it needs to get done

The timer thread is overly verbose and makes it difficult to see what craft-providers is doing in the other thread

mr-cal:
Related: https://github.com/canonical/craft-providers/blob/df23ecc1d8bb75304844d99bb74e3a89da9e2483/craft_providers/lxd/lxc.py#L1094 is also very verbose

PR #464, opened 2023-11-27
feat(lxd): store and check PID when launching instances

### LINKED ISSUE #448: Save and check PID during LXD base instance setup
### What needs to get done

If craft-providers fails during base setup, then subsequent runs fail after 60 seconds without a useful error message.  The only workaround is to manually delete the base instance.

### Why it needs to get done

Craft-providers should store the current PID in the lxd config of the base instance when setup begins.

If an existing base instance exists and is still in the PREPARING step, it should check the PID. If the PID is not currently running on the local machine, then the instance should be auto-cleaned (removed and recreated).

This may also require storing the name of the process as well, as the PID may not be unique across reboots.

### LINKED ISSUE #455: if an instance fails to launch, it can never be launched again
### Bug Description

If craft-providers fails to launch an instance [here](https://github.com/canonical/craft-providers/blob/df23ecc1d8bb75304844d99bb74e3a89da9e2483/craft_providers/lxd/lxc.py#L607), then it will create a scenario where the instance can never be launched again.

This happens because the instance is launched with `--config user.craft_providers.status=STARTING` and fails.  The next time craft-providers runs, it sees the instance is `STARTING`, wrongly assumes another process is starting the instance, and times out after 60 seconds.

To resolve this, craft-providers should delete the instance if it fails to launch.
https://github.com/canonical/craft-providers/issues/448 may also prevent this issue from occurring.

### To Reproduce

```python
from pathlib import Path
import logging

from craft_providers import bases, lxd

logging.basicConfig(level="DEBUG")
logger = logging.getLogger()

# define alias, provider, and base
alias = bases.BuilddBaseAlias.JAMMY
my_base = bases.BuilddBase(alias=alias)
provider_a = lxd.LXDProvider(lxd_project="project1")
provider_a.ensure_provider_is_available()
provider_b = lxd.LXDProvider(lxd_project="project2")
provider_b.ensure_provider_is_available()

with provider_a.launched_environment(
    project_name="hello-world",
    project_path=Path().absolute(),
    base_configuration=my_base,
    build_base=alias.value,
    instance_name="test-instance",
) as instance:
    logger.info("launched instance")

try:
    with provider_b.launched_environment(
        project_name="hello-world",
        project_path=Path().absolute(),
        base_configuration=my_base,
        build_base=alias.value,
        instance_name="test-instance",
    ) as instance:
        logger.info("launched instance")
except:
    logger.info("Failure is expected because of name conflict.")


# delete all instances in project1 to resolve name conflict
lxd.project.purge(lxc=lxd.LXC(), project="project1")

# the instance in project2 will never launch because its status is "STARTING"
with provider_b.launched_environment(
    project_name="hello-world",
    project_path=Path().absolute(),
    base_configuration=my_base,
    build_base=alias.value,
    instance_name="test-instance",
) as instance:
    logger.info("launched instance")
```

### part yaml

_No response_

### Relevant log output

```shell
n/a
```

mr-cal:

-----

The PID of the process that creates an instance is stored in the instance's LXD config and is used when launching an instance.

The `is_valid()` check for base instances now ensures an instance is stopped and has been fully setup.  If another process is busy launching and setting up the base instance, then `is_valid()` will wait (and eventually time out).  If the process that launched the base instance is inactive, then the instance is auto-cleaned (deleted and recreated).

This last statement has a slight behavioral change.  Previously, craft-providers raised a generic error (`Instance setup failed. Check LXD logs for more details`) and required the user to delete the base instance manually.  Now, the bad base instance will be auto-cleaned.  I think this is OK because this aligns with the design of the rest of `launcher` module where errors and misconfigured instances are auto-cleaned.

I recommend reviewing this per-commit.

The base instance has a new config value (the pid), so this must be paired with #465.

Fixes #448
Fixes #455 
(CRAFT-2221)

### DIFF
--- a/craft_providers/lxd/launcher.py
+++ b/craft_providers/lxd/launcher.py
@@ -25,6 +25,7 @@ import sys
 import threading
 import time
 from datetime import datetime, timedelta, timezone
+from pathlib import Path
 from typing import Optional
 
 from craft_providers import Base, ProviderError, bases
@@ -548,6 +549,80 @@ def _set_timezone(instance: LXDInstance, project: str, remote: str, lxc: LXC) ->
         value=timezone,
         project=project,
         remote=remote,
+    )
+
+
+def _wait_for_instance_ready(instance: LXDInstance) -> None:
+    """Wait for an instance to be ready.
+
+    If another process created an instance and is still running, then wait for
+    the other process to finish setting up the instance.
+
+    :param instance: LXD instance to wait for.
+
+    :raises LXDError:
+    - If the instance is not ready and the process that created the
+    instance is no longer running.
+    - If the function times out while waiting for another process
+    to set up the instance.
+    - On any failure to get config data or info from the instance.
+    """
+    instance_state = instance.info().get("Status")
+    instance_status = instance.config_get("user.craft_providers.status")
+
+    # check if the instance is ready
+    if (
+        instance_status == ProviderInstanceStatus.FINISHED.value
+        and instance_state == "STOPPED"
+    ):
+        logger.debug("Instance %r is ready.", instance.instance_name)
+        return
+
+    logger.debug(
+        "Instance %r is not ready (State: %s, Status: %s)",
+        instance.instance_name,
+        instance_state,
+        instance_status,
+    )
+
+    # LXD is linux-only, but verify the platform anyways
+    if sys.platform == "linux":
+        # get the PID of the process that created the instance
+        pid = instance.config_get("user.craft_providers.pid")
+
+        # an empty string means there was no value set
+        if pid == "":
+            raise LXDError(
+                brief="Instance is not ready.",
+                details=(
+                    f"Instance {instance.instance_name!r} is not ready and does not "
+                    "have the pid of the process that created the instance."
+                ),
+            )
+
+        # check if the PID is active
+        if Path(f"/proc/{pid}").exists():
+            logger.debug(
+                "Process that created the instance %r is still running (pid %s).",
+                instance.instance_name,
+                pid,
+            )
+        else:
+            raise LXDError(
+                brief="Instance is not ready.",
+                details=(
+                    f"Instance {instance.instance_name!r} is not ready and the process "
+                    f"(pid {pid}) that created the instance is inactive."
+                ),
+            )
+    else:
+        logger.debug("Skipping PID check because system is not linux")
+
+    # if the PID is active, wait for the instance to be ready
+    instance.lxc.check_instance_status(
+        instance_name=instance.instance_name,
+        project=instance.project,
+        remote=instance.remote,
     )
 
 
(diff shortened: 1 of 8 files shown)


craft_providers/lxd/launcher.py
```diff
@@ -551,6 +552,80 @@ def _set_timezone(instance: LXDInstance, project: str, remote: str, lxc: LXC) ->
...
+                ),
+            )
+
+        # check if the PID is active
+        if Path(f"/proc/{pid}").exists():
+            logger.debug(
+                "Process that created the instance %r is still running (pid %s).",
+                instance.instance_name,
+                pid,
+            )
```
syu-w, line 615: You may also want to check if `/proc/{pid}/exe` is symlink to a python.

craft_providers/lxd/launcher.py
```diff
@@ -264,35 +265,26 @@ def _formulate_base_instance_name(
...
-) -> bool:
+def _is_valid(*, instance: LXDInstance, expiration: timedelta) -> bool:
     """Check if an instance is valid.
 
-    Instances are valid if they are not expired (too old). An instance's age is measured
-    by it's creation date. For example, if the expiration is 90 days, then the instance
-    will expire 91 days after it was created.
+    Instances are valid if they are ready and not expired (too old).
+    An instance is ready if it is set up and stopped.
+    An instance's age is measured by it's creation date. For example, if the expiration
```
lengau, line 273:
```suggestion
    An instance's age is measured by its creation date. For example, if the expiration
```

craft_providers/lxd/lxc.py
```diff
@@ -1117,8 +1126,9 @@ def check_instance_status(
         # 20 * 3 seconds = 1 minute no change in timer
         timer_queue: deque = deque([-2, -1], maxlen=20)
 
-        # Retry unless the timer queue is all the same
+        # retry until the instance's timer hasn't changed for the last 20 iterations
         while len(set(timer_queue)) > 1:
+            logger.debug("Checking if instance is ready.")
```
lengau, line 1131: Not something to fix here, but are these debug logs the reason the cli timer resets every 3 seconds?

craft_providers/lxd/launcher.py
lengau: If we have permission to do so. For a process run by another user you won't be able to see the destination of the symlink unless you're root. (Easy test: set `pid` to 1, which should always be owned by root)

Although you can mostly get around that by reading the first string in `/proc/{pid}/cmdline`

lengau (approved):
LGTM

craft_providers/lxd/lxc.py
mr-cal: Correct.  I have that documented here as an enhancement:

https://github.com/canonical/craft-providers/issues/450#issuecomment-1811166186

craft_providers/lxd/launcher.py
mr-cal: I documented this enhancement here: https://github.com/canonical/craft-providers/issues/468

PR #482, opened 2023-12-11
[DRAFT] feat(buildd): use buildd images for mantic and devel

mr-cal:

-----

This PR is blocked by https://bugs.launchpad.net/cloud-images/+bug/2007419.

Once it is resolved, craft-providers should be able to use `buildd` images again for 23.10 and onwards.

I will also need to follow up with the Multipass team and see if we can launch `buildd` daily images via multipass.


mr-cal:
We can't use the `buildd` devel image because of #489.

The fix for Mantic is in SRU (https://bugs.launchpad.net/cloud-images/+bug/2007419).

When that lands, we can change this to only address Mantic (or just drop Mantic when it falls out of support in July-2024).

PR #532, opened 2024-03-14
fix(lxd): update LXD version naming support LTS

syu-w:


lengau (approved):
Nice, thanks!

sergiusens (approved):

sergiusens:
@mr-cal we might need an expedited craft-providers release

Issue #383, opened 2023-08-17
bumping the compatibility tag should be more robust


mr-cal:
### What needs to get done

Directly reference the compatibility tag from the tests, rather than hardcoding it.

### Why it needs to get done

When the compatibility tag needs to be bumped, it can be changed in 50 places in the test suite, but not all of them have to be changed.  This is confusing and can be error prone.

PR #549, opened 2024-04-11
Changes for 1.20.3

lengau:

-----

This is ready so we can do a rebase.

### DIFF
--- a/craft_providers/base.py
+++ b/craft_providers/base.py
@@ -979,6 +979,7 @@ class Base(ABC):
         *,
         executor: Executor,
         timeout: Optional[float] = TIMEOUT_UNPREDICTABLE,
+        mount_cache: bool = True,
     ) -> None:
         """Prepare base instance for use by the application.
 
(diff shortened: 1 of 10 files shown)


syu-w (approved):

mr-cal (commented):
What do you think about bumping the compatibility tag to `v8`?

`v7` is used by 1.23, so theoretically someone could create an instance with `1.20.3`, upgrade to `1.23.x` and the same instance would be reused (which we don't want).

I'm happy to do a compatibility tag bump commit, [it's more tedious than it should be](https://github.com/canonical/craft-providers/issues/383) and I've gotten pretty quick at it.

mr-cal:
> I've gotten pretty quick at it.

Meaning I run `git diff 69207f7~ 69207f7 --name-only | xargs sed -i 's/v6/v7/g'`

lengau:
Thanks @mr-cal  - I missed the v7 in 1.23. I've updated to v8 now

mr-cal (approved):
Thanks!

mr-cal:
Integration test failures are expected on this older branch. The tests are trying to launch 23.04 (lunar), which is no longer supported.

PR #562, opened 2024-04-25
docs: update to canonical-sphinx

tigarmo:
This one has multiple commits; the first one enables the extension and then I went a bit delete-happy and removed the sphinx-docs-starter-pack submodule (and references to it). My reasoning is that our repos are getting very confusing with regards to the docs - the starter pack has its own requirements files and other stuff that we no longer use, but it's hard to tell that at-a-glance.

I can drop the excess commits if we feel it's too much


lengau (commented):
This is fantastic, thanks!

PR #585, opened 2024-06-20
fix: check if lxd snap is installed

cmatsuoka:
Check if the lxd snap is installed by querying the snapd socket
instead of looking for an executable in the path, as lxd can use
auto-installer stubs.

-----

### DIFF
--- a/craft_providers/lxd/installer.py
+++ b/craft_providers/lxd/installer.py
@@ -19,6 +19,8 @@
 
 import logging
 import os
+import pathlib
+import shutil
 import subprocess
 import sys
 
@@ -107,6 +109,13 @@ def is_installed() -> bool:
     """
     logger.debug("Checking if LXD is installed.")
 
+    # check if non-snap lxd socket exists (for Arch or NixOS)
+    if (
+        pathlib.Path("/var/lib/lxd/unix.socket").is_socket()
+        and shutil.which("lxd") is not None
+    ):
+        return True
+
     # query snapd API
     url = "http+unix://%2Frun%2Fsnapd.socket/v2/snaps/lxd"
     try:
@@ -132,7 +141,7 @@ def is_installed() -> bool:
     logger.debug(f"LXD snap status: {status}")
     # snap status can be "installed" or "active" - "installed" revisions
     # are filtered from this API call with `select: enabled`
-    return bool(status == "active")
+    return bool(status == "active") and shutil.which("lxd") is not None
 
 
 def is_user_permitted() -> bool:
--- a/craft_providers/lxd/installer.py
+++ b/craft_providers/lxd/installer.py
@@ -110,9 +110,7 @@ def is_installed() -> bool:
     # query snapd API
     url = "http+unix://%2Frun%2Fsnapd.socket/v2/snaps/lxd"
     try:
-        snap_info = requests_unixsocket.get(
-            url=url, params={"select": "enabled"}
-        )
+        snap_info = requests_unixsocket.get(url=url, params={"select": "enabled"})
     except requests.exceptions.ConnectionError as error:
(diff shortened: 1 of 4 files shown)


craft_providers/lxd/installer.py
```diff
@@ -96,11 +101,42 @@ def is_initialized(*, remote: str, lxc: LXC) -> bool:
...
+        status = snap_info.json()["result"]["status"]
+    except KeyError:
+        raise errors.ProviderError(
+            brief="Unexpected response from snapd service."
+        )
+
+    logger.debug(f"LXD snap status: {status}")
+    # snap status can be "installed" or "active" - "installed" revisions
+    # are filtered from this API call with `select: enabled`
+    return bool(status == "active")
```
mr-cal, line 135: What do you think about retaining the old check as well?  It's unlikely to fail but may still be useful:
```
return bool(status == "active) and shutil.which("lxd") is not None
```

mr-cal (approved):
Looks good, nice unit test!

craft_providers/lxd/installer.py
cmatsuoka: @dariuszd21 also raised the issue about lxd on arch and nixos not being snap-packaged, we may need a dual check to cover these cases too.

craft_providers/lxd/installer.py
```diff
@@ -52,6 +54,7 @@ def install(sudo: bool = True) -> str:
 
     cmd += ["snap", "install", "lxd"]
 
+    logger.debug("installing lxd")
```
tigarmo, line 57:
```suggestion
    logger.debug("Installing LXD")
```

craft_providers/lxd/installer.py
```diff
@@ -62,6 +65,8 @@ def install(sudo: bool = True) -> str:
 
     lxd = LXD()
     lxd.wait_ready(sudo=sudo)
+
+    logger.debug("initializing lxd")
```
tigarmo, line 69:
```suggestion
    logger.debug("Initialising LXD")
```

tigarmo (approved):

craft_providers/lxd/installer.py
mr-cal: Yikes, that's a good point.

Does that stubbed LXD exec file on Ubuntu come in as an apt package?  Or is that just part of the base image?  

I'm wondering if craft-providers could check if LXD is installed as a snap only on Ubuntu with [`is_ubuntu_like`](https://github.com/canonical/craft-platforms/blob/a062dc03403c0404b09f6bb483b1151bb36d02f9/craft_platforms/_distro.py#L147).

Making an API call to LXC may be the most robust check.

cmatsuoka:
Re-requesting reviews because Dariusz raised a concern about a new scenario where non-snap LXD can be installed on the system.

craft_providers/lxd/installer.py
cmatsuoka: Covered this for now by verifying the presence of a non-snap lxd socket in the standard location, but a more robust verification would be always welcome.

sergiusens:
I am a bit conflicted by this as we can also be checking for `/var/snap/lxd/common/lxd/unix.socket` as it is their documented entrypoint https://github.com/canonical/lxd/blob/65de71d1fd90e71d84ad9bc238e9ea33bd2d1683/doc/howto/images_manage.md?plain=1#L295

cmatsuoka:
> I am a bit conflicted by this as we can also be checking for `/var/snap/lxd/common/lxd/unix.socket` as it is their documented entrypoint

By asking snapd we can have additional system health assurances (such as that snapd is up and running). Checking the presence of the socket file is what was possible in the non-snap case.

PR #602, opened 2024-07-22
fix(multipass): use Ubuntu 24.04 image

### LINKED ISSUE #511: Re-point multipass noble image when 24.04 is released
### What needs to get done

The only way to get a 24.04 buildd image with multipass is to call `multipass launch snapcraft:devel`.

When 24.04 is released, we need to update the MultipassProvider to call `multipass launch snapcraft:24.04`

### Why it needs to get done

To use the correct images

mr-cal:

-----

Use the buildd daily image for 24.04 for the Noble alias with Multipass

Fixes #511
Source: https://matrix.to/#/!GGqzbFAUQprdPgYYCM:ubuntu.com/$222ie8jdhmvzWbHqDyXE8QPgPtr4dUSph0J9T4pD6Xc?via=ubuntu.com&via=matrix.org&via=kde.org

### DIFF
--- a/craft_providers/multipass/multipass_provider.py
+++ b/craft_providers/multipass/multipass_provider.py
@@ -99,9 +99,8 @@ _BUILD_BASE_TO_MULTIPASS_REMOTE_IMAGE: Dict[Enum, RemoteImage] = {
     ubuntu.BuilddBaseAlias.MANTIC: RemoteImage(
         remote=Remote.RELEASE, image_name="mantic"
     ),
-    # this should use `image_name="24.04"` once noble is released (#511)
     ubuntu.BuilddBaseAlias.NOBLE: RemoteImage(
-        remote=Remote.SNAPCRAFT, image_name="devel"
+        remote=Remote.SNAPCRAFT, image_name="24.04"
     ),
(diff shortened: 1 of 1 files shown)


lengau (approved):

tigarmo (approved):

Saviq (commented):
Should there even be custom code like that? Shouldn't `build-base: devel` be required to build for a devel base?

mr-cal:
> Should there even be custom code like that? Shouldn't `build-base: devel` be required to build for a base in development?

You're correct, that would have been the correct path.  When we switched to core22, we made some design mistakes surrounding `base` and `build-base` that we are trying to rectify before core26 becomes the `devel` image.

mr-cal:
Ah, @lengau uncovered the original problem.  Mutlipass doesn't have a `snapcraft:core24` alias so we had been using `snapcraft:devel` to point to the 24.04 image.  This broke once 24.10 assumed the `devel` alias.

I should have asked the Multipass team about this a few months ago but I forgot: https://chat.canonical.com/canonical/pl/bxo9p7he37yt3kboepe9cp1y5e

PR #647, opened 2024-08-30
feat: acquire pro guest tokens in managed LXD instances

linostar:

-----

### DIFF
--- a/craft_providers/lxd/lxd_instance.py
+++ b/craft_providers/lxd/lxd_instance.py
@@ -19,6 +19,7 @@
 
 import hashlib
 import io
+import json
 import logging
 import os
 import pathlib
@@ -27,6 +28,8 @@ import shutil
 import subprocess
 import tempfile
 from typing import Any, Dict, List, Optional
+
+import requests
 
 from craft_providers.const import TIMEOUT_SIMPLE
 from craft_providers.errors import details_from_called_process_error
@@ -647,3 +650,41 @@ class LXDInstance(Executor):
             project=self.project,
             remote=self.remote,
         )
+
+    def retrieve_pro_host_token(self) -> str:
+        """Get the machine token from the pro host."""
+        try:
+            with open("/var/lib/ubuntu-advantage/private/machine-token.json") as fd:
+                content = json.load(fd)
+                return content.get("machineToken", "")
+        except (FileNotFoundError, PermissionError):
+            return ""
+
+    def request_pro_guest_token(self) -> str:
+        """Request a guest token from contracts API."""
+        machine_token = self.retrieve_pro_host_token()
+        if not machine_token:
+            # host is not pro
+            return ""
+
+        try:
+            base_url = "https://contracts.canonical.com"
+            endpoint = "/v1/guest/token"
+            response = requests.get(base_url + endpoint, headers={
+                "Authorization": f"Bearer {machine_token}"
+            })
+            if response.status_code != 200:
+                # fallback mechanism
+                return machine_token
+
+            guest_token = response.json().get("guestToken", "")
+            if not guest_token:
+                # fallback to machine token
+                return machine_token
+            return guest_token
+        except requests.exceptions.JSONDecodeError:
+            # recrived data was not in json format
+            return machine_token
+        except requests.exceptions.RequestException:
+            # guest token acquiring failed
+            return machine_token


linostar:
@canonical/starcraft-reviewers 
This is ready for review.

craft_providers/pro.py
```diff
@@ -0,0 +1,91 @@
...
+
+import requests
+
+from craft_providers.errors import MachineTokenError
+
+logger = logging.getLogger(__name__)
+
+
+class ProToken:
+    """Class for retrieving and processing pro tokens."""
```
tigarmo, line 30: Is there a specific reason to have a class here? Feels like it should just be a few free functions

craft_providers/pro.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @staticmethod
+    def retrieve_pro_host_token() -> str:
+        """Get the machine token from the pro host."""
+        try:
+            with open("/var/lib/ubuntu-advantage/private/machine-token.json") as fd:
```
tigarmo, line 36: why not use `Path.read_bytes()` instead?

craft_providers/errors.py
```diff
@@ -131,3 +131,7 @@ def __init__(self) -> None:
             f"see {url} for further reference."
         )
         super().__init__(brief=brief, resolution=resolution)
+
+
+class MachineTokenError(Exception):
```
tigarmo, line 136: this should be a subclass of `ProviderError`

craft_providers/pro.py
```diff
@@ -0,0 +1,91 @@
...
+                content = json.load(fd)
+                machine_token = content.get("machineToken", "")
+                if not machine_token:
+                    raise MachineTokenError("No token in machine token file.")
+                return machine_token
+        except FileNotFoundError:
+            raise MachineTokenError("Machine token file does not exist.")
+        except PermissionError:
+            raise MachineTokenError(
+                "Machine token file is not accessible. Make sure you are "
```
tigarmo, line 46: The second sentence should be the error's `resolution`

craft_providers/pro.py
```diff
@@ -0,0 +1,91 @@
...
+        try:
+            base_url = "https://contracts.canonical.com"
+            endpoint = "/v1/guest/token"
+            response = requests.get(
+                base_url + endpoint,
+                headers={"Authorization": f"Bearer {machine_token}"},
+                timeout=15,
+            )
+            if response.status_code != 200:
+                # fallback mechanism
```
tigarmo, line 64: what is this fallback mechanism here? I don't remember us talking about it, shouldn't it fail if we can't get a guest token?

tests/unit/test_pro.py
```diff
@@ -0,0 +1,158 @@
...
+from craft_providers.pro import ProToken
+from requests.exceptions import JSONDecodeError, RequestException
+
+_CONTRACTS_API_URL = "https://contracts.canonical.com"
+_CONTRACTS_API_ENDPOINT = "/v1/guest/token"
+
+
+@pytest.fixture
+def mock_machinetoken_open_success():
+    with mock.patch(
```
tigarmo, line 31: you can just use `mocker` instead, it's easier to read

```
def mock_machinetoken_open_success(mocker):
    return mocker.patch("builtins.open", new_callable=mock.mock_open,read_data=json.dumps({"machineToken": "test_token"}))

File: tests/unit/test_pro.py [hunk 7]
```diff
@@ -0,0 +1,158 @@
...
+def test_retrieve_pro_host_token(
+    fixture,
+    expected_output,
+    expected_exception,
+    expected_exception_message,
+    request,
+):
+    request.getfixturevalue(fixture)
+
+    if expected_exception:
```
tigarmo, line 105: I think this test is doing too much, and should be split into success/failure cases

tests/unit/test_pro.py
```diff
@@ -0,0 +1,158 @@
...
+def test_request_pro_guest_token(
+    mock_get,
+    mock_response_status,
+    mock_response_json,
+    mock_side_effect,
+    expected_token,
+    mock_retrieve_pro_host_token,
+):
+    """Test request_pro_guest_token behavior with various scenarios."""
+    if mock_side_effect:
```
tigarmo, line 146: same comment as the other test; this would be easier to read if it was split on multiple tests, each with a single intent

craft_providers/pro.py
```diff
@@ -0,0 +1,91 @@
...
+        except FileNotFoundError:
+            raise MachineTokenError("Machine token file does not exist.")
+        except PermissionError:
+            raise MachineTokenError(
+                "Machine token file is not accessible. Make sure you are "
+                "running with root access."
+            )
+
+    @staticmethod
+    def request_pro_guest_token() -> str:
```
tigarmo, line 51: I think we should add an integration test that mocks `retrieve_pro_host_token()` to a dummy value and then calls this, to smoke-test the logic of the code

tigarmo (changes requested):

craft_providers/pro.py
linostar: This is a temporary fallback mechanism because both the contracts API and the Pro backend's corresponding endpoint are not available yet. Once those are available we can remove this fallback.
For more info see https://warthogs.atlassian.net/browse/ROCKS-1460.

craft_providers/pro.py
linostar: Fixed

craft_providers/pro.py
linostar: I added a simple integration test for that function. Let me know if it's enough.

tigarmo (approved):

mr-cal (approved):

PR #656, opened 2024-09-19
feat: enabled pro services on managed LXD instance

linostar:

-----

This is a part of a series of adding Pro features to craft libraries.
This PR adds the functionality of enabling a single Pro service on a managed LXD instance.

### DIFF
--- a/craft_providers/lxd/lxc.py
+++ b/craft_providers/lxd/lxc.py
@@ -1218,3 +1218,46 @@ class LXC:
                 brief=f"Failed to run `pro` command on {instance_name!r}.",
                 details=errors.details_from_called_process_error(error),
             ) from error
+
+    def enable_pro_service(
+        self,
+        *,
+        instance_name: str,
+        service: str,
+        project: str = "default",
+        remote: str = "local",
+    ) -> None:
+        """Enable a Pro service on the instance.
+
+        :param instance_name: Name of instance.
+        :param service: Name of a single service to enable.
+        :param project: Name of LXD project.
+        :param remote: Name of LXD remote.
+
+        :raises LXDError: on unexpected error.
+        """
+        command = [
+            "pro",
+            "api",
+            f"{remote}:{instance_name}",
+            "u.pro.services.enable.v1",
+            "--data",
+            json.dumps({"service": service}),
+        ]
+        try:
+            self._run_lxc(
+                command,
+                capture_output=True,
+                project=project,
+            )
+
+            logger.debug(f"Pro service {service!r} successfully enabled on instance.")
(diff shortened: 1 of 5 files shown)


linostar:
@canonical/starcraft-reviewers

craft_providers/lxd/lxd_instance.py
```diff
@@ -660,3 +660,17 @@ def is_pro_enabled(self) -> bool:
             project=self.project,
             remote=self.remote,
         )
+
+    def enable_pro_service(self, service: str) -> None:
```
tigarmo, line 664: should probably be a `list[str]`, no?

linostar: The corresponding Pro API endpoint allows only for one service at a time.

tigarmo: right but if someone wants to enable multiple services, *some code* is going to have to loop and call the api multiple times, and it might as well be this api, no?

linostar: That will be handled on craft-application's side, because the list of services are passed there via `--pro` argument, so craft-application can call craft_providers.lxd.lxd_instance.enable_pro_service multiple times in a loop.

lengau: I agree with Tiago but with a slight change because I love abstractions:

```suggestion
    def enable_pro_services(self, services: Iterable[str]) -> None:
```

craft_providers/lxd/lxc.py
```diff
@@ -1218,3 +1218,51 @@ def is_pro_enabled(
                 brief=f"Failed to run `pro` command on {instance_name!r}.",
                 details=errors.details_from_called_process_error(error),
             ) from error
+
+    def enable_pro_service(
```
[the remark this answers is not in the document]
linostar: From my understanding, we won't need a `disable` method for the purpose of building Pro-enabed rocks, because of the following reasons:
- we are not going to auto-clean after finishing a pack or lifecycle command, so there will be no auto-detachment or auto-disable
- if `rockcraft clean` is run, the whole managed instance will be deleted anyway
- if `--pro` was passed to a subsequent lifecycle command with a different service(s), rockcraft will throw an error and stop (so it won't disabled the old service(s) and enable the new one(s))

craft_providers/lxd/lxd_instance.py
linostar: Changed the function signature so it accepts an iterable of `services` instead of a single `service`.

tigarmo (approved):

lengau (approved):

PR #708, opened 2025-01-06
ci: install lxd from candidate channel, install lxd on ubuntu 24.04

mr-cal:
- [ ] Have you followed the guidelines for contributing?
- [ ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] Have you successfully run `tox`?

-----

`ubuntu-latest` is being migrated to 24.04 this month.  Seems like LXD is no longer installed by default.

I took the liberty to test lxd on `candidate` while touching this.

source: https://github.com/canonical/craft-providers/pull/701

PR #710, opened 2025-01-07
feat: upstream snap hooks

mattculler:

-----

Applications using craft-providers may inadvertently leave around lxc instances, either when new base instances supersede old ones, or when the application is uninstalled.  Add a module implementing logic that applications can use in configure and remove snap hooks to prune these unneeded lxc instances.

### DIFF
--- a/craft_providers/hookutil.py
+++ b/craft_providers/hookutil.py
@@ -126,7 +126,7 @@ class HookHelper:
         # Didn't find our project name
         raise HookError(f"Project {self._project} does not exist in LXD.")
 
-    def dprint(self, *args: Any, **kwargs: Any) -> None:
+    def dprint(self, *args: Any, **kwargs: Any) -> None:  # noqa: ANN401
         """Print messages to stderr if debug=True.
 
         Can treat this like normal print(), except can also pass an instance
@@ -147,7 +147,7 @@ class HookHelper:
 
     def lxc(
         self,
-        *args: Any,
+        *args: Any,  # noqa: ANN401
         fail_msg: str | None = None,
         proj: bool = True,
         json_out: bool = True,
--- a/craft_providers/hookutil.py
+++ b/craft_providers/hookutil.py
@@ -50,7 +50,7 @@ class LXDInstance:
     name: str
     expanded_config: dict[str, str]
 
-    def base(self) -> str:
+    def base_instance_name(self) -> str:
         """Get the full name of the base instance this instance was created from."""
         try:
             return self.expanded_config["image.description"]
@@ -255,7 +255,7 @@ def configure_hook(lxc: HookHelper) -> None:
         # old (not future) and delete it.
         lxc.dprint(instance, "Base instance uses old compatibility tag")
         lxc.delete_instance(instance)
-        delete_base_full_names.add(instance.base())
+        delete_base_full_names.add(instance.base_instance_name())
 
     if not delete_base_full_names:
         lxc.dprint("No instances to delete")
@@ -263,7 +263,7 @@ def configure_hook(lxc: HookHelper) -> None:
 
     # Find the child instances of the bases we deleted and delete them too
     for instance in lxc.list_instances():
-        if instance.base() not in delete_base_full_names:
+        if instance.base_instance_name() not in delete_base_full_names:
             continue
         lxc.dprint("Base instance was deleted")
         lxc.delete_instance(instance)
--- a/craft_providers/hookutil.py
+++ b/craft_providers/hookutil.py
@@ -34,7 +34,6 @@ from typing_extensions import Self
 
 from craft_providers import Base
 
-
 _BASE_INSTANCE_START_STRING = "base-instance"
 _CURRENT_COMPATIBILITY_TAG_REGEX = re.compile(
     f"^{_BASE_INSTANCE_START_STRING}.*-{Base.compatibility_tag}-.*"
--- a/craft_providers/hookutil.py
+++ b/craft_providers/hookutil.py
@@ -67,7 +67,7 @@ class Instance:
         return self.name.startswith(_BASE_INSTANCE_START_STRING)
 
     @classmethod
-    def from_dict(cls, src: dict[str, str]) -> Self:
+    def unmarshal(cls, src: dict[str, str]) -> Self:
         """Use this rather than init - the lxc output has a lot of extra fields."""
         return cls(
             **{  # type: ignore[arg-type]
@@ -229,7 +229,7 @@ class HookHelper:
 
     def list_instances(self) -> list[Instance]:
         """Return a list of all instance objects for the project."""
-        return [Instance.from_dict(instance) for instance in self.lxc("list")]
+        return [Instance.unmarshal(instance) for instance in self.lxc("list")]
 
     def list_base_instances(self) -> list[Instance]:
         """Return a list of all base instance objects for the project."""
--- a/craft_providers/hookutil.py
+++ b/craft_providers/hookutil.py
@@ -25,13 +25,18 @@ From that, the thing we care most about is the compatibility tag:
 
 import dataclasses
 import json
+import re
 import subprocess
 import sys
 from typing import Any
 
 from typing_extensions import Self
 
+from craft_providers import Base
+
+
 _BASE_INSTANCE_START_STRING = "base-instance"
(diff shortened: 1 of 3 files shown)


mattculler:
Integration tests are failing due to ubuntu-latest switching to 24.04: https://github.com/canonical/craft-providers/pull/708

tests/unit/test_hookutil.py
mr-cal, line 1: Since craft-providers doesn't have spread, can you add an integration test that exercises `configure_hook()` and `remove_hook()`?

mr-cal (commented):
Looking good!

I'm not sure if the criteria for this task is to upstream or upstream and de-duplicate redundant code.    My comments are based on the latter.

The de-duplication is only possible now that this is internal to craft-providers and therefore can import other internal modules.

Either way, I would like to have an integration test for the configure and remove hooks.

craft_providers/hookutil.py
```diff
@@ -0,0 +1,280 @@
...
+    def is_current_base_instance(self) -> bool:
+        """Return true if this is a base instance with the current compat tag."""
+        return self.name.startswith(_CURRENT_BASE_INSTANCE_START_STRING)
+
+    def is_base_instance(self) -> bool:
+        """Return true if this is a base instance."""
+        return self.name.startswith(_BASE_INSTANCE_START_STRING)
+
+    @classmethod
+    def from_dict(cls, src: dict[str, str]) -> Self:
```
cmatsuoka, line 70: There's a precedent of calling this method `unmarshal` in pydantic models, should we use the pattern here as well?

craft_providers/hookutil.py
```diff
@@ -0,0 +1,280 @@
...
+                if k in {f.name for f in dataclasses.fields(cls)}
+            }
+        )
+
+
+class HookHelper:
+    """Hook business logic."""
+
+    def __init__(self, *, project_name: str, simulate: bool, debug: bool) -> None:
+        global _CURRENT_COMPATIBILITY_TAG, _CURRENT_BASE_INSTANCE_START_STRING  # noqa: PLW0603
```
cmatsuoka, line 85: Instead of using globals, can we have these only as instance attributes?

craft_providers/hookutil.py
```diff
@@ -0,0 +1,280 @@
...
+import dataclasses
+import json
+import subprocess
+import sys
+from typing import Any
+
+from typing_extensions import Self
+
+# These are interpolated once HookHelper is instantiated with the name of the project
+_CURRENT_COMPATIBILITY_TAG = "{PROJECT_NAME}-buildd-base-v7"
```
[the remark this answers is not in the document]
cmatsuoka: The way `_CURRENT_COMPATIBILITY_TAG` and `_CURRENT_BASE_INSTANCE_START_STRING` are stored seem unusual to me, can we have them obtained as properties of some other object instead of being globals? I think there are too many globals with long and similar names and that's confusing. I'd keep only the constants for magic values and assemble the others as needed (or set properties for them).

craft_providers/hookutil.py
```diff
@@ -0,0 +1,280 @@
...
+)
+
+
+class HookError(Exception):
+    """Hook logic cannot continue.  Hooks themselves should not exit nonzero."""
+
+
+@dataclasses.dataclass
+class Instance:
+    """Represents an lxc instance."""
```
cmatsuoka, line 48: If this is valid only for LXD instances, it can reflect it in the class name (`Instance` sounds too generic).

craft_providers/hookutil.py
```diff
@@ -0,0 +1,280 @@
...
+
+    name: str
+    expanded_config: dict[str, str]
+
+    def base(self) -> str:
```
cmatsuoka, line 53: Maybe this could be `base_instance_name` since `base` already has a different meaning.

craft_providers/hookutil.py
```diff
@@ -0,0 +1,280 @@
...
+        """Get the full name of the base instance this instance was created from."""
+        try:
+            return self.expanded_config["image.description"]
+        except KeyError as e:
+            # Unexpected, cannot continue
+            raise HookError("Could not get full base name from {self.name}") from e
+
+    def is_current_base_instance(self) -> bool:
+        """Return true if this is a base instance with the current compat tag."""
+        return self.name.startswith(_CURRENT_BASE_INSTANCE_START_STRING)
```
cmatsuoka, line 63: If `_CURRENT_BASE_INSTANCE_START_STRING` is used only here, you could hold the current compatibility tag as an instance attribute and assemble the start string inside this method.

cmatsuoka (changes requested):
The way the module data is organized is making the code hard to read, I think we can refactor this to make it easier for our future selves to rework this if needed. Also please add a description of the problem being solved to the PR so we can check if the implementation is doing everything that it's supposed to do.

craft_providers/hookutil.py
mattculler: Sure, I've renamed it.

craft_providers/hookutil.py
mattculler: Ok, renamed.

craft_providers/hookutil.py
mattculler: Ok, renamed

craft_providers/hookutil.py
mattculler: The reason I did this business with the globals was because it seemed less ugly to me than passing the application name to be stored as an instance variable for every `LXDInstance` object.  But it seems to have really thrown you for a loop, so I'll refactor the code to be more, for lack of a better word, _normal_.

mattculler:
> Looking good!
> 
> I'm not sure if the criteria for this task is to upstream or upstream and de-duplicate redundant code. My comments are based on the latter.
> 
> The de-duplication is only possible now that this is internal to craft-providers and therefore can import other internal modules.
> 
> Either way, I would like to have an integration test for the configure and remove hooks.

Awesome - based on the task pointage I was just going to upstream, not deduplicate.  I also didn't realize there would be so much deduplication possible - thanks for pointing all that out!  I'll open another ticket to capture your knowledge drop.

Re: integration tests, the private repo that this originated from has spread tests that do this, I'll see if I can rework those into integration tests here.

mr-cal (approved):
LGTM

craft_providers/hookutil.py
```diff
@@ -0,0 +1,280 @@
...
+        _CURRENT_BASE_INSTANCE_START_STRING = str.format(
+            _CURRENT_BASE_INSTANCE_START_STRING,
+            _BASE_INSTANCE_START_STRING=_BASE_INSTANCE_START_STRING,
+            _CURRENT_COMPATIBILITY_TAG=_CURRENT_COMPATIBILITY_TAG,
+        )
+
+        self._check_has_lxd()
+        self._check_project_exists()
+
+    def _check_has_lxd(self) -> None:
```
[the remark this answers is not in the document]
mattculler: Did this now because the snap check I had been using wouldn't work on non-snap systems (Nix, etc).

https://github.com/canonical/craft-providers/pull/710/commits/7fd4e24ac5c6d3ec683c6d31b60808ebc1b9ad5b

craft_providers/hookutil.py
```diff
@@ -0,0 +1,280 @@
...
+
+        print_args = list(args)
+        if len(args) >= 1 and isinstance(args[0], Instance):
+            # First arg quacks like an instance object
+            instance = print_args.pop(0)
+            print_args += [":", instance.name]
+
+        print(*print_args, **kwargs)
+
+    def lxc(
```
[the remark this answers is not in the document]
mattculler: https://github.com/canonical/craft-providers/issues/715

craft_providers/hookutil.py
```diff
@@ -0,0 +1,280 @@
...
+
+def remove_hook(lxc: HookHelper) -> None:
+    """Cleanup hook run on snap removal."""
+    for instance in lxc.list_instances():
+        lxc.delete_instance(instance)
+
+    # Project deletion will fail if images aren't all deleted first
+    lxc.delete_all_images()
+
+    lxc.delete_project()
```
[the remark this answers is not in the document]
mattculler: https://github.com/canonical/craft-providers/issues/715

cmatsuoka (approved):
Thanks! Remaining issues can be addressed in #715.

Issue #646, opened 2024-08-30
Switch from using the LXD CLI to pylxd


lengau:
### What needs to get done

Replace most usage of the `lxc` and `lxd` CLI with [pylxd](https://github.com/canonical/pylxd).

### Why it needs to get done

This is a faster, more reliable way to interact with LXD (e.g. https://github.com/canonical/craft-providers/issues/535)

mr-cal:
Here's the history: Craft-providers was written with the intent to support remote LXD servers.  `pylxd` didn't support that when craft-providers was created, so we created our own abstraction layer around LXD.

lengau:
Given our own slow progress on remote LXD instances, I wonder if helping `pylxd` to support remotes would be a reasonable way to do it.

steinbro:
We should revisit this when we migrate from LXD 5.x to 6.x. The latter is set to get its LTS release this year per https://discourse.ubuntu.com/t/lxd-6-3-has-been-released/56974

Issue #535, opened 2024-03-15
Make LXD version check more robust


mr-cal:
### What needs to get done

If LXD version parsing fails, craft-providers should catch the failure and log a warning.

### Why it needs to get done

LXD recently changed their version naming convention, which broke craft-providers and had to be fixed via #532.  This check should be more robust.

lengau:
From the LXD team:

> The CLI output isn't guaranteed to be stable, so perhaps a better way is to use the API directly, i.e.

```
lxc query /1.0 | jq .environment.server_version
"5.21.1"
```

> The API version server_version didn't change, and instead we added a new server_lts field. 

so querying the API directly is the right way to go

simondeziel:
Just an FYI, that currently there is a [bug in `lxd-installer`](https://bugs.launchpad.net/ubuntu/+source/lxd-installer/+bug/2069754) (a kind of shim to dynamically install LXD's snap if it's not present) that prevents the suggested detection mechanism from working.

Until that bug fix is propagated to all supported releases, please make sure to test on systems where the LXD snap is installed rather than rely on `lxd-installer` to pull it.

PR #720, opened 2025-01-28
feat: fail if users try to run managed builds that would use a nonfunctional container

mattculler:

-----

New LXD versions were released that fix an issue we've seen where a mismatch between the support for cgroupsv1 and v2 between host and guest causes the container's systemd to become partially nonfunctional.  Add checks to fail with a good error message if a user's system would launch such a container.

(CRAFT-3096)


craft_providers/lxd/lxc.py
```diff
@@ -1173,3 +1176,25 @@ def check_instance_status(
             time.sleep(3)
 
         raise LXDError(brief="Timed out waiting for instance to be ready.")
+
+    def get_version(self) -> str:
```
mr-cal, line 1180: Can you use `craft_providers.lxd.lxd.version`?

craft_providers/bases/ubuntu.py
```diff
@@ -271,3 +279,59 @@ def _clean_up(self, executor: Executor) -> None:
...
+    return BuilddBaseAlias(version_id)
+
+
+def ensure_guest_compatible(base_configuration: Base, guest_instance: Executor) -> None:
+    """Ensure host is compatible with guest instance."""
+    if not issubclass(base_configuration, BuilddBase):
+        # Not ubuntu, not sure how to check
+        return
+
+    host_instance = LoopbackExecutor()
```
mr-cal, line 297: What happens on non-Ubuntu hosts?

mr-cal (commented):
Looks good overall.  Seems possible to get this done with or without the loopback executor.  I'd like to get @lengau's opinion on that, since he was talking about this when architecting craft-application.

craft_providers/bases/ubuntu.py
mattculler: Unknown - it would depend on the version of systemd, kernel version, and possibly also how the kernel was compiled.

mattculler:
> Looks good overall. Seems possible to get this done with or without the loopback executor. I'd like to get @lengau's opinion on that, since he was talking about this when architecting craft-application.

Yeah you could.  Taking a fresh look at it today I see a good way to refactor `Base._get_os_release` so the meat of it it reusable without `Executor` - going to try that instead.

mattculler:
> > Looks good overall. Seems possible to get this done with or without the loopback executor. I'd like to get @lengau's opinion on that, since he was talking about this when architecting craft-application.
> 
> Yeah you could. Taking a fresh look at it today I see a good way to refactor `Base._get_os_release` so the meat of it it reusable without `Executor` - going to try that instead.

I think this is better, going this route and removing loopback executor unless you disagree: https://github.com/canonical/craft-providers/pull/720/commits/d2ff08628bbee2678efb79392a009d67471223b5

craft_providers/lxd/lxc.py
mattculler: Good call, definitely can, thanks for pointing this out

mr-cal:
> @mr-cal I think this is better, going this route and removing loopback executor unless you disagree: [d2ff086](https://github.com/canonical/craft-providers/commit/d2ff08628bbee2678efb79392a009d67471223b5)

I like it.  I'm interested in a loopback, but keeping this PR lean will speed up the dev and review process.

craft_providers/lxd/lxc.py
mattculler: https://github.com/canonical/craft-providers/pull/720/commits/5878afca0f52ff411bcadf028057e8ec29cd994b

Only unknown is whether the permissions will work out correctly - I think calls to lxd need root.

craft_providers/bases/ubuntu.py
```diff
@@ -271,3 +279,64 @@ def _clean_up(self, executor: Executor) -> None:
...
+    if lxd_major == 5:
+        # Major is 5, we care about patch versions given the minor
+        if lxd_minor == 0 and lxd_patch < 4:
+            raise lxd_exception
+        if lxd_minor == 21 and lxd_patch < 2:
+            raise lxd_exception
+    if lxd_major < 5:
+        raise lxd_exception
+
+    kernel_version = subprocess.run(
```
lengau, line 332: This should use [`platform.release`](https://docs.python.org/3/library/platform.html#platform.release). We might want to check that we're on Linux for it too, as other OS's can have `lxd` that just passes to a remote server.

craft_providers/bases/ubuntu.py
```diff
@@ -271,3 +279,64 @@ def _clean_up(self, executor: Executor) -> None:
...
+        # LXD version strings sometimes omit the patch - call it zero
+        lxd_patch = 0
+    lxd_exception = ProviderError(
+        brief="This combination of guest and host OS versions requires a newer lxd version.",
+        resolution="Ensure you have lxd >=5.0.4 or >=5.21.2 installed - try the lxd snap.",
+    )
+    if lxd_major == 5:
+        # Major is 5, we care about patch versions given the minor
+        if lxd_minor == 0 and lxd_patch < 4:
+            raise lxd_exception
```
lengau, line 326: We can detect both the lxd version issue and the kernel version issue in one run and raise both problems simultaneously.

craft_providers/lxd/lxc.py
```diff
@@ -1173,3 +1176,29 @@ def check_instance_status(
...
+
+    def get_server_version(self) -> str:
+        """Get the version of the lxd server.
+
+        Use over LXD.version when you need to get the server version but aren't root.
+
+        :raises LXDError: on unexpected error.
+        """
+        try:
+            completed_process = self._run_lxc(
```
lengau, line 1188: If we already have infrastructure for calling the lxd [REST API](https://documentation.ubuntu.com/lxd/en/latest/rest-api/) we should use that instead. (I don't remember whether we do.)

craft_providers/lxd/lxd_provider.py
```diff
@@ -161,6 +161,15 @@ def launched_environment(
         except BaseConfigurationError as error:
             raise LXDError(str(error)) from error
 
+        # It would be ideal to check if the instance is compatible before creating/launching
+        # it, but unfortunately image_name can be a lot of things, so we have to read the OS
+        # release from the running instance.
```
lengau, line 166: In that case we should be able to check directly on the systemd version rather than being dependent on the Ubuntu version, right?

craft_providers/bases/ubuntu.py
```diff
@@ -29,13 +30,16 @@
...
+    ProviderError,
     details_from_called_process_error,
 )
 from craft_providers.executor import Executor
+from craft_providers.util.os_release import parse_os_release
 
 logger = logging.getLogger(__name__)
 
 
+@total_ordering
```
lengau, line 42: Nice use of total_ordering here!

craft_providers/lxd/lxc.py
mr-cal: We don't yet (maybe you are thinking of https://github.com/canonical/craft-providers/issues/535#issuecomment-2049860524), but we could lean on `pylxd` for calls like this instead of API calls (#646) in the future.

craft_providers/bases/ubuntu.py
```diff
@@ -271,3 +279,64 @@ def _clean_up(self, executor: Executor) -> None:
...
+        brief="This combination of guest and host OS versions requires a newer lxd version.",
+        resolution="Ensure you have lxd >=5.0.4 or >=5.21.2 installed - try the lxd snap.",
+    )
+    if lxd_major == 5:
+        # Major is 5, we care about patch versions given the minor
+        if lxd_minor == 0 and lxd_patch < 4:
+            raise lxd_exception
+        if lxd_minor == 21 and lxd_patch < 2:
+            raise lxd_exception
+    if lxd_major < 5:
```
mr-cal, line 329:
```suggestion
    elif lxd_major < 5:
```

craft_providers/bases/ubuntu.py
```diff
@@ -271,3 +279,64 @@ def _clean_up(self, executor: Executor) -> None:
...
+    # Without loopback executor:
+    host_base_alias = BuilddBaseAlias(parse_os_release().get("VERSION_ID"))
+
+    # If the host OS is focal (20.04) or older, and the guest OS is oracular (24.10)
+    # or newer, then the host lxd must be >=5.0.4 or >=5.21.2, and kernel must be
+    # 5.15 or newer.  Otherwise, weird systemd failures will occur due to a mismatch
+    # between cgroupv1 and v2 support.
+    # https://discourse.ubuntu.com/t/lxd-5-0-4-lts-has-been-released/49681#p-123331-support-for-ubuntu-oracular-containers-on-cgroupv2-hosts
+    if (
+        host_base_alias > BuilddBaseAlias.FOCAL
```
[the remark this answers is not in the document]
mr-cal: Agreed, this design would let us make changes in the future without needing to refactor.

mr-cal (commented):
Looks good overall. Please re-request me after addressing Alex's comments

craft_providers/bases/ubuntu.py
mattculler: Good call on platform.release.  Though re: checking we're on linux, by this point we've already verified that our base OS is an affected version.

craft_providers/lxd/lxc.py
mattculler: Using the lxc REST API is a good idea, could you open a craft-providers ticket for that?

craft_providers/lxd/lxd_provider.py
mattculler: Hypothetically yes.  We could also hypothetically check for the specific cgroup version the kernels have support for.  But I've implemented the checking based on what the lxd team described their bug fix as applying to - namely, host and guest ubuntu versions/lxd version/host kernel version.

craft_providers/bases/ubuntu.py
mattculler: Thanks!  It's a fantastic little decorator.

craft_providers/bases/ubuntu.py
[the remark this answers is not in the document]
mattculler: *Not exactly like that, I'm refining this design as I implement, but this is the gist.

craft_providers/bases/ubuntu.py
mattculler: Yeah, this is probably the pattern I'll move towards.  The error messages will be a little less specific, but I  think that's an acceptable tradeoff here.

craft_providers/bases/ubuntu.py
mattculler: This logic will go away with the generic design Alex suggested above.

craft_providers/bases/ubuntu.py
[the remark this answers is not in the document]
mr-cal: Yeah, this is getting complex fast.  Seems like we need a class or library to handle versioning more cleanly.  I'd be fine to leave is as you originally implemented.

Also, this code may not be long-lived.  It's going to be increasingly rare for someone to use these older minor versions of lxd.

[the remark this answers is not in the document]
mattculler: @mr-cal The code I've got right now is working out a little more cleanly than the above would suggest, mainly by making some non-generic assumptions about lxd versions.  I wasn't sold on this approach before, but I think what I'm settling on now is better - look for an update shortly.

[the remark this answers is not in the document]
lengau: Could we forget about 5.0.4 and just check if the version tuple `>= (5, 21, 2)`?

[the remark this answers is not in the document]
mr-cal: @lengau, what's your rationale for 5.21 and not 5.0? Both are LTS versions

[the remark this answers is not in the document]
mattculler: These two commits implement a generic solution:
https://github.com/canonical/craft-providers/pull/720/commits/f4c8161e739f2ebf2f401b87f0802b15f86ad830
https://github.com/canonical/craft-providers/pull/720/commits/31dce457af374e26f7cc88aebbae862e2a2332c7

But, I'm torn - I feel like the generic way it's harder to reason about what the code is doing, and we're unlikely to need to add to this INVALID_VERSIONS structure.  Thoughts? @lengau @mr-cal ?

craft_providers/bases/checks.py
```diff
@@ -0,0 +1,138 @@
...
+    BIONIC = "18.04"
+    FOCAL = "20.04"
+    JAMMY = "22.04"
+    NOBLE = "24.04"
+    ORACULAR = "24.10"
+    DEVEL = "devel"
+"""
+
+
+INVALID_VERSIONS = [
```
mr-cal, line 56: Nice, this is very readable

mr-cal (commented):
I'm +1 for your new design, readable and seems simpler since it separates this particular cgroups bug away from the if/else statements.

craft_providers/bases/checks.py
```diff
@@ -0,0 +1,138 @@
...
+    ProviderError,
+    details_from_called_process_error,
+)
+from craft_providers.executor import Executor
+from craft_providers.util.os_release import parse_os_release
+
+logger = logging.getLogger(__name__)
+
+
+"""
```
lengau, line 45: What's this docstring for?

tests/unit/bases/test_ensure_guest_compatible.py
```diff
@@ -0,0 +1,155 @@
...
+    fake_get_os_release.counter = 0  # type: ignore[reportFunctionMemberAccess]
+    guest_base._get_os_release = fake_get_os_release
+
+    lxd_version = "0.0.0"
+
+    # Mock the host os-release file
+    fake_os_release = f"""
+VERSION_ID="{base_alias.value}"
+WOOP="dedoo"
+"""
```
lengau, line 72:
```suggestion
    fake_os_release = textwrap.dedent(f"""\
        VERSION_ID="{base_alias.value}"
        WOOP="dedoo"
        """
    )
```

craft_providers/bases/checks.py
```diff
@@ -0,0 +1,138 @@
...
+INVALID_VERSIONS = [
+    {
+        "host_less_than_equal": BuilddBaseAlias.FOCAL,
+        "guest_greater_than_equal": BuilddBaseAlias.ORACULAR,
+        # The system is affected by the cgroups bug if both of the above and either of the below
+        "lxd_less_than": [
+            (5, 0, 4),
+            (5, 21, 2),
+        ],
+        "kernel_less_than": (5, 15),
```
lengau, line 65: How about making this a list of triples as well and using the same function for matching lxd and kernels?

craft_providers/bases/checks.py
mattculler: For my reference, didn't mean to commit it there on Friday.  Removed.

craft_providers/bases/ubuntu.py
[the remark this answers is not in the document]
lengau: @mr-cal my rationale is that anything after 5.21.2 works, but some things greater than 5.0.4 don't work.

craft_providers/bases/checks.py
lengau: It's not necessary now, but it could be necessary in the future, especially given OEM kernels.

lengau (approved):

mr-cal (approved):

PR #767, opened 2025-06-30
feat: optionally delay shutdown in a launched environment.

### LINKED ISSUE #706: Delay the stopping of instances when requested.
### What needs to get done

Basically make https://github.com/canonical/craft-providers/pull/571 production-ready.

- Change `Provider.launched_environment()` to take an optional shutdown delay time (default to None using current behaviour, otherwise delay for the specified number of minutes).
- Change each `Provider` to be able to handle an already-running instance. (Perhaps setting and releasing a lock on that instance?)
- Change each `Provider` to run the `shutdown` command if the instance has a shutdown delay time set (even if that time is 0, in which case the shutdown should still occur asynchronously).

### Why it needs to get done

During iteration on a (snap|charm|rock|whatever), each run includes starting up and shutting down an instance, which can add significantly to the time the run takes. A pre-warmed machine can avoid this.

NOTE: It will be up to the application to actually opt into using this feature, so it will be a backwards compatible change.

lengau:
CRAFT-4377

Fixes #706

---

### DIFF
--- a/craft_providers/lxd/lxd_instance.py
+++ b/craft_providers/lxd/lxd_instance.py
@@ -32,7 +32,10 @@ from craft_providers.errors import details_from_called_process_error
 from craft_providers.executor import Executor, get_instance_name
 from craft_providers.lxd.errors import LXDError
 from craft_providers.lxd.lxc import LXC
-from craft_providers.lxd.lxd_instance_status import LXDInstanceState
+from craft_providers.lxd.lxd_instance_status import (
+    LXDInstanceState,
+    ProviderInstanceStatus,
+)
 from craft_providers.util import env_cmd, retry
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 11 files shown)


craft_providers/lxd/lxd_instance.py
```diff
@@ -556,11 +577,31 @@ def _is_running(timeout: float) -> None:
             error=LXDError(brief="Instance failed to restart."),
         )
 
-    def stop(self) -> None:
+    def stop(self, delay_mins: int = 0) -> None:
```
lengau, line 580: Per conversation with @mr-cal:

Default this to `None` and if it's any integer including zero run the new code path

craft_providers/lxd/lxd_instance.py
```diff
@@ -513,6 +516,22 @@ def start(self) -> None:
...
         logger.info("Starting instance")
+        if self.info().get("Status") == LXDInstanceState.RUNNING.value:
+            if (
+                state := self.config_get("user.craft_providers.status")
+            ) == ProviderInstanceStatus.FINISHED.value:
+                logger.debug("Instance already running but available.")
+                self.config_set(
+                    "user.craft_providers.status", ProviderInstanceStatus.IN_USE.value
+                )
+                self.execute_run(["shutdown", "-c"])
```
cmatsuoka, line 527: We could hide this in a `cancel_shutdown()` method or similar, since the command is called more than once.

craft_providers/lxd/lxd_instance.py
```diff
@@ -556,11 +578,31 @@ def _is_running(timeout: float) -> None:
...
+
+        If delay_mins is 0 (the default), this method waits for the shutdown to finish.
+        If it's nonzero, the method returns after preparing the shutdown.
         """
+        if delay_mins is not None:
+            logger.debug(f"Shutting down after {delay_mins} minutes")
+            self.config_set(
+                "user.craft_providers.status", ProviderInstanceStatus.FINISHED.value
+            )
+            self.execute_run(
```
cmatsuoka, line 595: And this could be a `shutdown(delay_mins)`.

cmatsuoka (approved):
This seems to implement the functionality we need.

craft_providers/lxd/lxd_instance.py
lengau: I'm concerned about adding these methods because we probably don't want to make it part of the public API. How about if I add them but make them private?

PR #788, opened 2025-08-11
feat: prepare instance before base setup

cmatsuoka:
Introduce a mechanism to prepare the instance after creation and before
any files are downloaded. This allows the application to set up HTTP
proxies and CA certificates to be used in the download process.

---
CRAFT-4679


craft_providers/lxd/launcher.py
```diff
@@ -147,6 +154,9 @@ def _create_instance(  # noqa: PLR0913
             config_timer = InstanceTimer(base_instance)
             config_timer.start()
 
+            if prepare_instance:
+                prepare_instance(instance)
```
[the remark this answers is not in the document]
lengau: @cmatsuoka I think copilot is right here. The next thing we do is run `setup` on the base instance, but if the base instance isn't prepared like this its setup will fail if it's required to be behind the proxy.

Another possibility: if we're setting a `prepare_instance` callback like this, maybe we should never have a base instance?

tigarmo (commented):
I think the change looks alright design-wise, but I think we need to see the craft-application integration working before we approve and merge this

craft_providers/lxd/launcher.py
```diff
@@ -95,6 +101,7 @@ def _create_instance(  # noqa: PLR0913
     project: str,
     remote: str,
     lxc: LXC,
+    prepare_instance: "Callable[[Executor], None] | None" = None,
```
mr-cal, line 104: Can you add a `:param:` in the docstring?

edit: This is in a few more functions throughout the library - do you mind updating the docstrings for all of them? I won't leave half a dozen of the same comments.

craft_providers/multipass/multipass_provider.py
```diff
@@ -170,7 +176,7 @@ def ensure_provider_is_available(cls) -> None:
             install()
         ensure_multipass_is_ready()
 
-    def create_environment(self, *, instance_name: str) -> Executor:
+    def create_environment(self, *, instance_name: str) -> "Executor":
```
mr-cal, line 179: nit: if you do a `from __future__ import annotations`, then you don't have to do this legacy practice of quoting types.

mr-cal (approved):

craft_providers/lxd/launcher.py
cmatsuoka: Added missing params.

PR #810, opened 2025-08-23
feat(lxd): more robust version parsing

### LINKED ISSUE #535: Make LXD version check more robust

lengau:
This makes the LXD version check more robust by doing the following:

1. Both the LXC and LXD versions first try using the API.
2. If that fails they fall back to the existing command-based version.

`LXD.is_supported_version` assumes the version is supported if it cannot be parsed (for example, edge versions with version "git-[commit]")

Fixes #535
CRAFT-2619

### DIFF
--- a/craft_providers/lxd/lxd.py
+++ b/craft_providers/lxd/lxd.py
@@ -22,8 +22,8 @@ import pathlib
 import subprocess
 
 import packaging.version
-import pylxd
-from pylxd.exceptions import ClientConnectionFailed
+import pylxd  # type: ignore[import-untyped]
+from pylxd.exceptions import ClientConnectionFailed  # type: ignore[import-untyped]
 
 from craft_providers.errors import details_from_called_process_error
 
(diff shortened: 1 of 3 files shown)


craft_providers/lxd/lxd.py
```diff
@@ -76,28 +83,49 @@ def is_supported_version(self) -> bool:
...
 
     def version(self) -> str:
         """Query LXD version.
 
         The version is of the format:
-        <major>.<minor>[.<micro>] [LTS]
+        <major>.<minor>[.<micro>]
 
         Version examples:
-        - 5.21.0 LTS
```
mr-cal, line 94: Couldn't this format get returned if it falls back to `return self._version_cmd()`?

lengau: Yes, good point. I was originally going to just use pylxd, but if the lxd command is installed but lxd isn't running that'd be a regression.

PR #885, opened 2025-12-17
style: enable type checking with ty

lengau:

---


craft_providers/multipass/multipass_instance.py
bepri: The return types around this and `execute_run` seem odd. I feel like it should be able to discern whether or not it is `Popen` or `CompletedProcess`, with `[str]` or `[bytes]`, as I believe LXD already does. Have a go at it, but if it isn't reasonably simple then I say just make an issue and merge this.

Issue #783, opened 2025-08-04
Snap info should use a Pydantic model


bepri:
### What needs to get done

In the below code, we have to just assume that we'll get the desired output from deserializing JSON:
https://github.com/canonical/craft-providers/blob/5717e2c0907c2bb304d0d3752671be68e9b1fa14/craft_providers/actions/snap_installer.py#L134

This should instead make use of a Pydantic model to validate and type-hint the response from the snapd API.

### Why it needs to get done

A Pydantic model allows type checkers to be used during development for the values returned and enables runtime sanity checks on snapd's response.

PR #908, opened 2026-02-12
feat(models): add SnapInfo model for snapd response

PraaneshSelvaraj:
Converted Snap Info to a Pydantic model and added test cases.

Fixes: #783 

---

### DIFF
--- a/craft_providers/models/__init__.py
+++ b/craft_providers/models/__init__.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2021-2023 Canonical Ltd.
+# Copyright 2026 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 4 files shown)


bepri:
`extra="ignore"` is the default value, which seems to be what we use in other places such as the Craft Store library models. Let's stick with that.

For consistency's sake, I think what makes the most sense is to model the set of fields in the example response you found in the test. That one seems to be a good set of fields that we use today, plus the functional ones I can imagine us possibly using in the future.

Thanks :)

bepri:
Thanks, @PraaneshSelvaraj! The code looks great, definitely a huge improvement.

Could you take a look at the failing workflows [here](https://github.com/canonical/craft-providers/actions/runs/22248186277/job/64543904249?pr=908) and [here](https://github.com/canonical/craft-providers/actions/runs/22248186277/job/64543904286?pr=908)? They are related to your changes but should be easy fixes :)

PR #951, opened 2026-04-22
fix(Base): log the output of successful commands as debug.

### LINKED ISSUE #250: bases: subprocess output is not logged nor shown on error
`craft-providers` does not capture and display the output of subprocesses.

Two related issues:
1. When running with trace-level logging, the output of the commands should be logged.
    - This is how snapcraft core18|20 operates.
3. When a subprocess command fails, the command's output should be shown.
    - For example, when [`apt-get update` returns exit code 100](https://forum.snapcraft.io/t/snapcraft-network-error-seemingly-with-all-core22/34588), the output of the command would be very useful for debugging.

lengau:
## Reproducer for #250

craft-providers runs many commands inside instances (`apt-get`, `systemctl`, `snap`, etc.) but **never logs their output**, even when running with debug/trace logging. Operators diagnosing problems have to reproduce a full failure to see what a command printed.

### Failing test

**`test_execute_run_logs_output_at_debug`** — shows that `_execute_run()` captures stdout via `capture_output=True` but never writes it to the logger. After the command completes successfully, the output is simply discarded.

### Bonus passing test

**`test_execute_run_output_included_in_error_details`** — verifies that the *other* half of #250 ("output not shown on error") **already works**: when a subprocess fails, its stdout is preserved on the `CalledProcessError` and included in the `BaseConfigurationError` details via `details_from_called_process_error()`. This serves as a regression guard.

### Fix

In `Base._execute_run()`, after `executor.execute_run()` returns successfully, log `proc.stdout` and `proc.stderr` at `logger.debug` level (guarded by a truthiness check so empty output is not logged).

Closes #250

### DIFF
--- a/tests/unit/bases/test_ubuntu_buildd.py
+++ b/tests/unit/bases/test_ubuntu_buildd.py
@@ -1674,6 +1674,67 @@ def test_execute_run_bad_check_verifynetwork_combination(fake_executor):
         )
 
 
+def test_execute_run_logs_output_at_debug(fake_executor, fake_process, logs):
+    """_execute_run() should log command stdout at debug level.
+
+    When craft-providers runs commands inside an instance (apt-get, systemctl,
+    etc.) the output is currently silently discarded on success.  Operators
+    running with debug/trace logging expect to see that output so they can
+    diagnose problems without having to reproduce a full failure.
+
+    Regression test for https://github.com/canonical/craft-providers/issues/250
+    """
+    base_config = ubuntu.BuilddBase(alias=ubuntu.BuilddBaseAlias.JAMMY)
+    command = ["the", "command"]
+    fake_process.register_subprocess(
+        [*DEFAULT_FAKE_CMD, *command],
+        returncode=0,
+        stdout=b"important command output",
+    )
+
+    base_config._execute_run(command, executor=fake_executor)
+
+    assert "important command output" in logs.debug
+
+
+def test_execute_run_output_included_in_error_details(fake_executor, fake_process):
+    """When a subprocess fails, its stdout must appear in the error details.
+
+    For example, when `apt-get update` returns exit code 100, the apt output
+    contains the specific mirror/package that caused the failure.  Without it
+    the user sees only "Failed to update apt cache" with no actionable detail.
+
+    The test registers a subprocess that produces stdout and fails, then
+    checks that a higher-level caller wraps the error with those details
+    visible — using _pre_setup_packages (apt-get update) as a representative
+    caller.
+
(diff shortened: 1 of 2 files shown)


tigarmo:
how noisy is this on a regular run?

bepri:
I like the change, but I second Tiago's question -- if it's particularly noisy, I might suggest moving this to trace instead.

lengau:
@tigarmo @bepri logger.debug does only get output by craft-cli in trace mode.

bepri (approved):

tests/unit/bases/test_ubuntu_buildd.py
```diff
@@ -1674,6 +1674,67 @@ def test_execute_run_bad_check_verifynetwork_combination(fake_executor):
...
 
+def test_execute_run_logs_output_at_debug(fake_executor, fake_process, logs):
+    """_execute_run() should log command stdout at debug level.
+
+    When craft-providers runs commands inside an instance (apt-get, systemctl,
+    etc.) the output is currently silently discarded on success.  Operators
+    running with debug/trace logging expect to see that output so they can
+    diagnose problems without having to reproduce a full failure.
+
+    Regression test for https://github.com/canonical/craft-providers/issues/250
```
mr-cal, line 1680: I think the test purpose and link to the issue is perfectz:

```suggestion
    """_execute_run() should log command stdout at debug level.

    Regression test for https://github.com/canonical/craft-providers/issues/250
```

tests/unit/bases/test_ubuntu_buildd.py
```diff
@@ -1674,6 +1674,67 @@ def test_execute_run_bad_check_verifynetwork_combination(fake_executor):
...
+    For example, when `apt-get update` returns exit code 100, the apt output
+    contains the specific mirror/package that caused the failure.  Without it
+    the user sees only "Failed to update apt cache" with no actionable detail.
+
+    The test registers a subprocess that produces stdout and fails, then
+    checks that a higher-level caller wraps the error with those details
+    visible — using _pre_setup_packages (apt-get update) as a representative
+    caller.
+
+    Regression test for https://github.com/canonical/craft-providers/issues/250
```
mr-cal, line 1698:
```suggestion
    """When a subprocess fails, its stdout must appear in the error details.

    Regression test for https://github.com/canonical/craft-providers/issues/250
```

mr-cal (commented):
This will be a very helpful improvement, just being critical on readability.

mr-cal (approved):
Thanks!

cmatsuoka:
@lengau a manual main merge is required, Github refuses to do that due to "Repository rule violations found".

PR #973, opened 2026-06-04
chore: merge the latest Starbase

lengau:

---


docs/explanation/index.rst
```diff
@@ -1,8 +1,14 @@
+.. meta::
+    :description: The explanatory documentation for Craft Providers. These explanations aid in understanding the concepts and relationships of Craft Providers as a complete system.
```
jahn-junior, line 2:
```suggestion
    :description: These explanations aid in understanding the concepts and relationships of Craft Providers as a complete system.
```
What do you think of dropping the first sentences from these meta descriptions? The first sentence feels redundant given the link text that these will accompany.

Image

jahn-junior (approved):
Thanks! The docs changes LGTM.

One suggestion regarding meta descriptions, but I don't consider it a blocker for library docs.

docs/explanation/index.rst
medubelko: This comes from Starbase itself. Let's upstream any fixes rather than discuss it here.

bepri (approved):

docs/explanation/index.rst
jahn-junior: Oops. I didn't know we added these upstream.

lengau: I'm happy to do that, but since this is just a Starbase merge I'm going to do so in a separate PR.

